PR #21, opened 2014-10-30
Fix error message in case of unknown criterion value

krocard:
When failing to parse a criterion rule, the pfw was outputing: `Value error:`
This did not help to find the root cause.

With this patch it would have output:
`Value error: <Input value> is not part of criterion <Criterion>`
(Of course replacing <*> by the apropriate value)


parameter/SelectionCriterionRule.cpp
```diff
@@ -105,7 +105,8 @@ bool CSelectionCriterionRule::parse(CRuleParser& ruleParser, string& strError)
     // Value
     if (!_pSelectionCriterion->getCriterionType()->getNumericalValue(strValue, _iMatchValue)) {
 
-        strError = "Value error: " + strError;
+        strError = "Value error: " + strValue + " is not part of criterion " +
```
dawagner: Please add quotes around the string values

PR #32, opened 2014-12-19
Remove explicit server stop

krocard:
Stopping the remote-processor in the stop command has two side effect:
 - stopping the remote-processor thread will fail as it is not possible
   to join a thread to itself. This leads to the leak of all the thread
   object (destructor not called).
 - the server will not return the exit command status as it has been
   stopped during it's execution.

Remove the remote-processor stop as it will be called during it's
destruction.

### DIFF
--- a/test/test-platform/TestPlatform.cpp
+++ b/test/test-platform/TestPlatform.cpp
@@ -154,9 +154,6 @@ CTestPlatform::CommandReturn CTestPlatform::exit(
 {
     (void)remoteCommand;
 
-    // Stop local server
-    _pRemoteProcessorServer->stop();
-
     // Release the main blocking semaphore to quit application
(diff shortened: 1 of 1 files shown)


krocard:
Rebased to trigger a build from travis.

krocard:
Anyone ? It is very easy to reproduce:

# Without the patch
~~~ {cpp}
$ test-platform config.xml &
$ remote-process localhost 5001 exit
Unable to received answer from target: Size read failed: Connection reset by peer

# $? == 1; 2 valgrind leak
~~~
# With the patch
~~~ {cpp}
$ test-platform config.xml &
$ remote-process localhost 5001 exit

# $? == 0; 0 valgrind leak
~~~

PR #33, opened 2014-12-19
Handle config path without slash

krocard:
If the parameter configuration file path does not contain a "/"
it used to assert as it could not find it's folder.

Use "." this case.


Makohoek:
:+1:

krocard:
:+1: @gdenneul @dawagner

krocard:
@dawagner review

PR #35, opened 2015-01-02
test-platform: fix boolean remote getter commands output

dawagner:
Getter commands return a boolean value ("true" or "false") were implemented by
a generic method.  However, this method returned a "Done" result, which
suppresses the normal output and prints "Done" instead.

This is fixed by making this method return a "Success" result, which is
interpreted as "give the output to the user".

In the same time, we remove a useless "cast to void" which is usually a
workaround for unused variables.

Change-Id: I49115d3b29967fd455ffc0b703dfb4f4d438280e


dawagner:
@krocard please review

krocard:
:+1:

PR #40, opened 2015-02-04
Remove CXX compiler verbose flag for swig binding

krocard:
When compiling swig generated files, the CXX compiler
flag list included -v. This leaded to unnecessary log
output during compilation. =>_Remove the flag.

Some cmake command were in upper case.


Makohoek:
:+1:

dawagner:
:-1: It breaks:

    swig error : Unrecognized option -Wall -Werror

see http://stackoverflow.com/questions/17043088/cmake-build-calling-swig-with-multiple-arguments

The correct version would be:

    SET_PROPERTY(SOURCE pfw.i PROPERTY SWIG_FLAGS "-Wall" "-Werror")

and the current makefile is wrong too.

krocard:
:-1:

krocard:
:+1:

PR #41, opened 2015-02-04
Fix pointer arithmetic instead of concat

krocard:
`"my number " + 5` does not produces `"my number 5"` but rather `"mber "`.
If the number was greater than `len(string) + 2`
(pointing to the \0 and one past the end in authorized, more in UB)
the result would have been undefined behaviour.

Concatenate the number with a string stream.


krocard:
:+1: 
@dawagner @Makohoek

PR #44, opened 2015-02-06
Missing std prefix in skeleton plugin

krocard:
The skeleton plugin is not compiled when compiling
the root pfw project. As a result, when `using namespaces`
was removed from the pfw headers the skeleton plugin
was left unchecked and unfortunately broken.

Fix the missing `std::` by adding them where needed or
inserting a `using std::string`.

Signed-off-by: Kevin Rocard <kevin.rocard@intel.com>

### DIFF
--- a/skeleton-subsystem/SkeletonSubsystem.h
+++ b/skeleton-subsystem/SkeletonSubsystem.h
@@ -34,7 +34,7 @@
 class CSkeletonSubsystem : public CSubsystem
 {
 public:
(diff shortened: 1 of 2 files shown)


krocard:
:+1:  @dawagner @Makohoek

Makohoek:
:+1:

Makohoek:
You could add "this patch fixs issue #44" to benefit from githubs "auto close issue feature"

dawagner:
see #47

PR #45, opened 2015-02-10
Use minidom factory methods in xml coverage generator

krocard:
The exportToXML was creating an minidom.Element using it's constructor, which is forbidden in the documentation:
> Applications should not instantiate the classes themselves; they should use the creator functions available on the Document object.
http://docs.python.org/library/xml.dom.minidom.html

Since python 3.4 this method no longer works (private interface change) thus the xml coverage generation was broken.

Fix by calling the recommended method Document.createElement.

See bug: http://bugs.python.org/issue15290

Signed-off-by: Kevin Rocard <kevin.rocard@intel.com>

### DIFF
--- a/tools/coverage/coverage.py
+++ b/tools/coverage/coverage.py
@@ -204,14 +204,14 @@ class Element():
                 str(dumpedDescription) for dumpedDescription in
                         self._dumpDescription(withCoverage, withNbUse))
 
-    def exportToXML(self, domElement=None):
+    def exportToXML(self, document, domElement=None):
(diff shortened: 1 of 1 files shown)


krocard:
@gdenneul @Makohoek @dawagner  :+1:

Makohoek:
Commit seems good to me, however the renaming could be done in another patch even if it makes sense to rename it here since you introduce the "document" variable in this patch.
:+1:

dawagner:
:+1:

PR #47, opened 2015-02-13
missing std in skeleton

dawagner:


dawagner:
:+1: 
@krocard I added a commit on top of your pull request

PR #48, opened 2015-02-16
Introduce remote-process bash completion

krocard:
When working with the parameter-framework,
the remote-process command is very handy.
Nevertheless it has long commands and it is easy to make mistakes.

Add a bash completion script to complete:
 - hostnames
 - frequently used port
 - commands and arguments based on the remote usage (help command)


krocard:
:+1:

krocard:
:+1:

PR #53, opened 2015-02-23
C pfw bindings

krocard:
The pfw can not currently be used from c code.

Add an c api. It does not target a perfect one/one mapping with the c++ one,
but rather aim ease of use and type safety (as far as possible in c).
This api is unit tested (100% coverage).

Enable automatic unit test run with coverage in travis.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,7 +12,7 @@ before_install:
   - sudo apt-get update -qq
 
 install:
-  - sudo apt-get install --yes swig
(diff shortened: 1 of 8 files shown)


krocard:
@sguiriec @Makohoek @sguiriec @dawagner @gdenneul please review.

krocard:
The pull request was splited to #61and #63 to integreat it faster.

PR #55, opened 2015-02-25
Parameter manager multithreading usage

tcahuzax:
The libxml2 library, which is a dependency of
the parameter framework, requires a specific usage
in case of multithreading.

Indeed, if the libxml2 library is used by several
threads, the function xmlInitParser() has to be called
at program start in the main thread (i.e. a thread that
lives until process exit).

This patch introduces a static method "initForMultithreading"
to follow this requirement. The user has to call it at program
start in the main thread if multithreading is used.

Signed-off-by: Thomas Cahuzac <thomasx.cahuzac@intel.com>


dawagner:
See #246 instead.

PR #57, opened 2015-03-03
Let a domain decide when he should be synchronized.

clero:
Domains are browsed twice when applying configurations.
The first one gather syncer of domains which are not
"sequence aware" and launch a sync on them.
The second one launch an apply without providing a syncer set.
It lets sequence aware domains to synchronize if needed.

This patch allows to browse domains only once. If a domain
is sequence aware, it will be synchronized during configurations
restauration. If not, the synchronisation will be delayed at
the end of all applies.

Sequence aware domains are now synchronized first.


clero:
:+1: @krocard @dawagner

krocard:
:+1:

benaware:
:-1: Synchronization of all sequence unaware domains needs to say coalesced. This is because the synchronization of sequence aware domains is susceptible to do some read/modify/write operations on syncers which are shared with sequence unaware ones. By synchronizing sequcence unware ones first, we ensure a proper initialization of the shared syncers.

PR #58, opened 2015-03-05
Log rework

clero:
Log rework to remove them for Element class.
It will require plugins update.


clero:
:+1:  @Makohoek @krocard @dawagner @OznOg

clero:
This pull request conflicts with c bindings as those bindings assume that no status can be returned in case of a success.
This pull request introduces an information retrieval possibility for the client when a criterion is set. This information allows the client to gather some information about criterion applications (number of change since last apply).
Maybe the solution is to give the possibility to set a list a criterion, the apply would be useless and so will be the information.

clero:
Add Android compilation support :+1:

dawagner:
merged into the 'clang' branch.

clero:
rebased on functional tests

parameter/ConfigurableDomain.h
```diff
@@ -89,14 +112,31 @@ class CConfigurableDomain : public CBinarySerializableElement
...
-    // Domain splitting
-    bool split(CConfigurableElement* pConfigurableElement, std::string& strError);
+    /** Split the domain in two.
+     * Remove an element of a domain and create a new domain which owns the element.
+     *
+     * @param[in] pConfigurableElement pointer to the element to remove
+     * @param[out] infos useful information we can provide to client
+     * @return true if succeed false otherwise
+     *
+     * FIXME: You may want to remove me
```
dawagner: not related to this pull request. Please remove & feel free to open a ticket.

parameter/ConfigurableDomain.h
```diff
@@ -148,12 +188,26 @@ class CConfigurableDomain : public CBinarySerializableElement
...
+
     void mergeConfigurations(CConfigurableElement* pToConfigurableElement, CConfigurableElement* pFromConfigurableElement);
 
-    // Configurable elements association
-    void doAddConfigurableElement(CConfigurableElement* pConfigurableElement, const CParameterBlackboard* pMainBlackboard = NULL);
+    /** Actually realize the association between the domain and a configurable  element
+     *
+     * @param[in] pConfigurableElement pointer to the element to add
+     * @param[out] infos useful information we can provide to client
+     * @param[in] pMainBlackboard pointer to the application main blackboard
```
dawagner: Please explain what is supposed to happen if pMainBlackboard == NULL. If there is a way to tell doxygen that the argument is optional, please use it (but maybe doxygen already knows this by parsing the prototype).

parameter/ConfigurableDomains.cpp
```diff
@@ -93,8 +98,12 @@ void CConfigurableDomains::apply(CParameterBlackboard* pParameterBlackboard, CSy
 
         const CConfigurableDomain* pChildConfigurableDomain = static_cast<const CConfigurableDomain*>(getChild(uiChild));
 
+        std::string info;
         // Apply and synchronize when relevant
-        pChildConfigurableDomain->apply(pParameterBlackboard, NULL, bForce);
+        pChildConfigurableDomain->apply(pParameterBlackboard, NULL, bForce, info);
+        if (info.length() != 0) {
```
dawagner: `if (!info.empty())`

parameter/ConfigurableDomains.h
```diff
@@ -79,7 +79,21 @@ class CConfigurableDomains : public CBinarySerializableElement
...
+
+    /** Split a domain in two.
+     * Remove an element of a domain and create a new domain which owns the element.
+     *
+     * @param[in] domainName the domain name
+     * @param[in] element pointer to the element to remove
+     * @param[out] infos useful information we can provide to client
+     * @return true if succeed false otherwise
+     *
+     * FIXME: You may want to remove me
```
dawagner: same as above

parameter/ConfigurableDomains.h
```diff
@@ -88,7 +102,22 @@ class CConfigurableDomains : public CBinarySerializableElement
...
-    bool restoreConfiguration(const std::string& strDomain, const std::string& strConfiguration, CParameterBlackboard* pMainBlackboard, bool bAutoSync, std::list<std::string>& lstrError) const;
+
+    /** Restore a configuration
+     *
+     * @param[in] strDomain the domain name
+     * @param[in] strConfiguration the configuration name
+     * @param[in] mainBlackboard the application main blackboard
+     * @param[in] autoSync boolean which indicates if auto sync mechanism is on
+     * @param[out] errors, errors encountered during restoration
+     * @result true is success false otherwise
```
dawagner: "result" and "return" are equivalent but "return" is used everywhere else.

parameter/DefaultElementLibrary.h
```diff
@@ -46,7 +47,13 @@ class CDefaultElementLibrary: public CElementLibrary
...
 public:
 
-    explicit CDefaultElementLibrary(bool bEnableDefaultMechanism = true);
+    /**
+     * Class constructor
+     *
+     * @param[in] logger the logger provided by the client
+     * @param[in] bEnableDefaultMechanism boolean allowing to activate VirtualSubsystem use
+     */
+    explicit CDefaultElementLibrary(core::log::ILogger& logger, bool bEnableDefaultMechanism = true);
```
dawagner: Although it does no harm keeping the `explicit` keyword, I'm wondering (for my own curiosity): is it still needed ? This constructor is no longer a single-argument ctor and can therefore no longer be used for implicit conversion (even though it has an optional argument ?), right ?

Anyway, the 'explicit' keyword should probably be kept here.

parameter/DefaultElementLibrary.h
```diff
@@ -71,15 +78,29 @@ class CDefaultElementLibrary: public CElementLibrary
       */
     CElement* createElement(const CXmlElement& xmlElement) const;
 
+    /**
+     * Application logger getter
+     *
+     * FIXME: Avoid the use of this getter by reworking plugin builder declaration.
```
dawagner: Do it or open a ticket.

parameter/DefaultElementLibrary.h
```diff
@@ -71,15 +78,29 @@ class CDefaultElementLibrary: public CElementLibrary
...
+     * @return the logger provided by the client
+     */
+    core::log::ILogger& getLogger();
+
 private:
     bool _bEnableDefaultMechanism;
     CDefaultElementBuilder _DefaultElementBuilder;
+
+    /** Application Logger we need to provide to the plugin */
+    core::log::ILogger& _Logger;
```
dawagner: s/_Logger/_logger/

parameter/DomainConfiguration.h
```diff
@@ -69,8 +70,18 @@ class CDomainConfiguration : public CBinarySerializableElement
...
     void save(const CParameterBlackboard* pMainBlackboard);
-    // Apply data to current
-    bool restore(CParameterBlackboard* pMainBlackboard, bool bSync, std::list<std::string>* plstrError = NULL) const;
+
+    /** Restore the configuration
+     *
+     * @param[in] pMainBlackboard the application main blackboard
+     * @param[in] bSync indicates if a synchronisation has to be done
+     * @param[out] errors, errors encountered during restoration
+     * @result true is success false otherwise
```
dawagner: same as above

parameter/ConfigurableDomains.h
dawagner: You have the same typo in lots of other places

parameter/ParameterMgr.cpp
```diff
@@ -102,6 +98,9 @@ using std::list;
 using std::vector;
 using std::ostringstream;
 
+// FIXME: integrate ParameterMgr to core namespace
```
dawagner: do it ?

parameter/ParameterMgr.cpp
```diff
@@ -308,8 +307,8 @@ CParameterMgr::CParameterMgr(const string& strConfigurationFilePath) :
     _uiStructureChecksum(0),
     _pRemoteProcessorServer(NULL),
     _uiMaxCommandUsageLength(0),
-    _pLogger(NULL),
-    _uiLogDepth(0),
+    _Logger(logger),
```
dawagner: _logger

parameter/ParameterMgr.cpp
```diff
@@ -308,8 +307,8 @@ CParameterMgr::CParameterMgr(const string& strConfigurationFilePath) :
...
+    _Prolog(""),
```
dawagner: _prolog

parameter/log/include/log/Logger.h
```diff
@@ -0,0 +1,47 @@
...
+{
+
+/** Default information logger type */
+typedef details::LogWrapper<true, Context> Info;
+
+/** Default warning logger type */
+typedef details::LogWrapper<false, Context> Warning;
+
+} /** log namespace */
+} /** core namespace */
```
dawagner: This file doesn't seem to be used by anyone.

parameter/DefaultElementLibrary.h
clero: Why? Current coding style uses a capital letter after the underscore like 
~~~c++
CDefaultElementBuilder _DefaultElementBuilder;
~~~
or this is a mistake too ?

parameter/DefaultElementLibrary.h
clero: you're right it has no effect. I forgot to remove it, but as you said, it can be good to keep it anyway

parameter/DefaultElementLibrary.h
dawagner: That one is a mistake too, indeed.

parameter/DefaultElementLibrary.h
dawagner: yup, keept it :)

parameter/ConfigurableDomain.h
clero: You are only talking about the FIXME right?

dawagner: right (although one could argue that the documentation could first be added in a separate patch and then updated when adding the 'infos' argument ;)).

clero: done

parameter/ConfigurableDomain.h
clero: apparently there is not, doxygen advise is to document it explicitly

parameter/ConfigurableDomains.cpp
clero: done

parameter/ConfigurableDomains.h
clero: done

parameter/ConfigurableDomains.h
clero: done

parameter/ParameterMgr.cpp
clero: it will lead to unrelated changes, i will open a ticket and do it in an other pull request

parameter/log/include/log/Logger.h
clero: I forgot to remove it

parameter/CMakeLists.txt
```diff
@@ -1,4 +1,4 @@
-# Copyright (c) 2014, Intel Corporation
+# Copyright (c) 2015, Intel Corporation
```
dawagner: 2014-2015

parameter/ConfigurableDomain.h
```diff
@@ -148,12 +186,28 @@ class CConfigurableDomain : public CBinarySerializableElement
...
+
     void mergeConfigurations(CConfigurableElement* pToConfigurableElement, CConfigurableElement* pFromConfigurableElement);
 
-    // Configurable elements association
-    void doAddConfigurableElement(CConfigurableElement* pConfigurableElement, const CParameterBlackboard* pMainBlackboard = NULL);
+    /** Actually realize the association between the domain and a configurable  element
+     *
+     * @param[in] pConfigurableElement pointer to the element to add
+     * @param[out] infos useful information we can provide to client
+     * @param[in] (optionnal) pMainBlackboard, pointer to the application main blackboard
```
dawagner: s/optionnal/optional/

parameter/ParameterMgr.cpp
```diff
@@ -93,15 +90,27 @@
...
-#include <memory>
 
 #define base CElement
 
+/** Private macro helper to declare a new context
+ *
+ * Context declaration always need logger and logging prefix to be
+ * passed as parameters.
+ * This macro aims to avoid this boring notation.
+ * This macro should be called only once in a scope.
```
dawagner: This is not completely accurate. Consider the following code:

```
{
  // Scope a
  LOG_CONTEXT(...);

  {
    // Scope b
    LOG_CONTEXT(..);
  } // End of scope b
} // End of scope a
```

This code is fine even though the macro is called twice in Scope `a`. However, I'm not sure how to phrase this... Maybe "This macro may only be called once in a scope's root but may be called again in nested scopes." ?

parameter/DefaultElementLibrary.h
krocard: No it is still needed
~~~
struct I { I(int a, int b=0) {} };
struct E { explicit E(int a, int b=0) {} };
struct F { static void i(I) {}
               static void e(E) {} };
F::i(1); // OK
F::e(1); // Error: no viable conversion from 'int' to 'E'
~~~
Note: I wrapped i(I) and e(E) in a class F because cling does not allow function definition out of a class.

parameter/ParameterMgr.cpp
```diff
@@ -941,7 +940,7 @@ CParameterMgr::CCommandHandler::CommandStatus CParameterMgr::statusCommandProces
 
     /// Criteria states
     appendTitle(strResult, "Selection Criteria:");
-    list<string> lstrSelectionCriteria;
+    core::Results lstrSelectionCriteria;
```
dawagner: in most places, you've removed hungarian notation when using "Results". you should do it here too.

parameter/ParameterMgr.cpp
clero: as a new scope is defined I suppose it's understandable (it was what I meant). Maybe we can precise that variable shadowing is authorized..

parameter/DefaultElementLibrary.h
krocard: Shorter (same) example:
~~~ {cpp}
struct I1 {          I1(int a=0       ) {} };
struct I2 {          I2(int a, int b=0) {} };
struct E1 { explicit E1(int a=0       ) {} };
struct E2 { explicit E2(int a, int b=0) {} };

template <class T> void f(T) {};

f<I1>();  // Error: too few arguments to function call, expected 1, have 0
f<I1>(1); // OK
f<I2>(1); // OK

f<E1>();  // Error: too few arguments to function call, expected 1, have 0
f<E1>(1); // Error: no viable conversion from 'int' to 'E'
f<E2>(1); // Error: no viable conversion from 'int' to 'E'
~~~

dawagner:
Please rebase against the master branch (conflicts are minor).

parameter/DefaultElementLibrary.h
clero: done

parameter/CMakeLists.txt
clero: done

parameter/ParameterMgr.cpp
clero: done

parameter/ConfigurableDomain.h
clero: done

clero:
rebase done

parameter/ParameterMgr.cpp
clero: done in #87

dawagner:
replaced by #87

PR #62, opened 2015-03-13
C parameter framework bindings

krocard:
The pfw can not currently be used from c code.

Add an c api. It does not target a perfect one/one mapping with the c++ one,
but rather aim ease of use and type safety (as far as possible in c).
This api is unit tested (100% coverage).


krocard:
Except the last commit, the commits are from #61.

dawagner:
:+1:

PR #63, opened 2015-03-13
Fixed point test rework

dawagner:
The fixed point test was not returning any error code which made it seem as though it was successful...

In addition to this fix, the test suite now uses the python bindings instead of the remote interface which makes them much faster to execute.

However, some tests are failing in some cases (some incorrect values with high precision are accepted as valid by the parameter-framework).


dawagner:
The failures mentioned in the PR message will need further investigation.

@krocard please review

dawagner:
@krocard please review

dawagner:
@Makohoek please review

PR #65, opened 2015-03-16
travis: set the installation prefix to /usr

dawagner:
CMake defaults to /usr/local which results in libraries being installed in
/usr/local/lib. Unfortunately, this path isn't in the loader's default
locations.

The canonical installation path is /usr. /usr/local is recommended for stuff
not handled by the package manager but in the context of the CI, this isn't
relevant.


dawagner:
@krocard please review

PR #72, opened 2015-03-19
Tokenizer: unit tests and new implementation

dawagner:


utility/Tokenizer.cpp
```diff
@@ -1,125 +1,75 @@
...
+
+string Tokenizer::next()
+{
+    string token;
+
+    // Skip all leading delimiters
+    string::size_type tokenStart = _input.find_first_not_of(_delimiters, _position);
+
+    // Special case if there isn't any token anymore (string::substr's
+    // behaviour doesn't seem to be defined when pos==npos)
```
krocard:
~~~ {cpp}
basic_string substr( size_type pos = 0,
                     size_type count = npos ) const;
~~~
if `count` == `npos`, the returned substring is [pos, size()).
Source: http://en.cppreference.com/w/cpp/string/basic_string/substr

utility/Tokenizer.cpp
```diff
@@ -1,125 +1,75 @@
...
+    vector<string> result;
+    string token;
+
+    while (true) {
+        token = next();
+        if (token.empty()) {
+            return result;
+        }
+        result.push_back(token);
+    }
```
krocard:
~~~ {cpp}
for (string token = next(); not token.empty; token = next())
    result.push_back(token);
~~~
return result;

utility/Tokenizer.h
```diff
@@ -1,56 +1,66 @@
...
-
-    std::string buffer;                                 // input string
-    std::string token;                                  // output string
-    std::string delimiter;                              // delimiter string
-    std::string::const_iterator currPos;                // string iterator pointing the current position
-
-};
-
-#endif // TOKENIZER_H
+
```
krocard: class doxygen please.

utility/Tokenizer.h
```diff
@@ -1,56 +1,66 @@
...
+     * "a     bc d   " will be tokenized as ("a", "bc", "d") if the delimiter
+     * is ' '.
+     */
+    std::string next();
+
+    /** Return a vector of all tokens
+     */
+    std::vector<std::string> split();
+
+    static const std::string defaultDelimiters;
```
krocard: Explain what are the default delimiters, it is very important for the client.

PR #73, opened 2015-03-20
Install c parameter-framework bindings

krocard:
C bindings were compiled and tested but never installed.

Add the missing install command.


krocard:
@Makohoek @dawagner @gdenneul

Makohoek:
:+1:

PR #75, opened 2015-03-27
Log/error message improvement/fix in the xml coverage generator

krocard:
Improve and fix various error/log messages of the python report generator python script.

### DIFF
--- a/tools/coverage/coverage.py
+++ b/tools/coverage/coverage.py
@@ -59,11 +59,11 @@ class ChildError(CustomError):
 
 class ChildNotFoundError(ChildError):
     def __str__(self):
-        return "Unable to find the child %s in %s" % (self.child, self.parent)
+        return 'Unable to find the child "%s" in "%s"' % (self.child, self.parent)
 
 class DuplicatedChildError(ChildError):
     def __str__(self):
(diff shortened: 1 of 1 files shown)


tools/coverage/coverage.py
```diff
@@ -834,7 +834,7 @@ def _digest(self, lineLogType, lineLog):
 
 
     def parsePFWlog(self, lines):
-        for lineNb, lineLog in enumerate(lines):
+        for lineIndex, lineLog in enumerate(lines):
```
dawagner: The offset may be given here:

    enumerate(lines, 1)
    # or
    enumerate(lines, start=1)

see https://docs.python.org/3/library/functions.html#enumerate

dawagner:
We should add some unit tests for this coverage script, one of these days (same for the edd parser and xml generator and espacially those, since we use them more often).

tools/coverage/coverage.py
krocard: Good point

krocard:
@dawagner A first easy step would be to generate an output given an input and compare with a reference. As those scripts are stateless, it should cover most of the code.

PR #80, opened 2015-04-07
Fail memchek in case of leak or error

krocard:
Some unit test were checked using valgrind, but if an error was detected it was only logged, the build was not failed.

Fail to build all *MemCheck targets if a leak or an error is reported by valgrind.

Also add cleanup of test related files and cmakelist


krocard:
:+1: @dawagner @clero @Makohoek

CMakeLists.txt
```diff
@@ -54,6 +54,13 @@ add_subdirectory(remote-processor)
 
 add_subdirectory(remote-process)
 
+# todo: Can those valgrind flags be moved in CTestCustom.cmake ?
```
gdenneul: This is worse trying, no? Isn't it strange to have tests here at the base level?

valgrind.supp
```diff
@@ -0,0 +1,11 @@
...
+# It is fixed in zlib1g v1.2.8 (ubuntu 14.04)
+# but not in v1.2.3 (ubuntu 12.04) which travis has.
+{
+   libz/inflateReset2/jump
+   Memcheck:Cond
+   fun:inflateReset2
+   fun:inflateInit2_
+   obj:/*/libz.so*
+}
+
```
dawagner: Please put this file somewhere else (i.e. not at the repository's root).

CMakeLists.txt
krocard: I tried rely hard but did not succeed. It must be done before the `include(CTest)` the only thing I can do is to create a repertory ctest and include all ctest related things inside. That seems a good idea, isn't it ?

valgrind.supp
krocard: Agree

krocard:

dawagner:
:+1:

PR #88, opened 2015-04-11
Improve travis debug build

krocard:
- Avoid running `gcov` twice.
- Run `make test` *without* `make install` in debug.
- Fix coveralls reporting headers with 0 code as 0% covered instead of 100%.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,6 +21,10 @@ install:
 
 # how to build
 script:
+  - ( mkdir build_debug && cd build_debug &&
+        cmake -DCMAKE_BUILD_TYPE=Debug -DCOVERAGE=ON .. &&
+        make -j &&
+        CTEST_OUTPUT_ON_FAILURE=1 make ExperimentalTest ExperimentalMemCheck )
   - ( mkdir build && cd build &&
         cmake .. &&
         make -j &&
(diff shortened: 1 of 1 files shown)


krocard:
> Fix coveralls reporting headers with 0 code as 0% covered instead of 100%.

Strangely *parameter/include/ParameterMgrFullConnector.h* is still reported as 0% covered although coveralls says that it has 0 "relevant lines". Compare:
https://coveralls.io/builds/2318278/source?filename=parameter%2Finclude%2FParameterMgrFullConnector.h#L52 (this pull request)
https://coveralls.io/builds/2309517/source?filename=parameter%2Finclude%2FParameterMgrFullConnector.h#L52 (master)

krocard:

.travis.yml
```diff
@@ -32,19 +36,14 @@ script:
         make &&
         sudo make install &&
         sudo ldconfig )
```
dawagner: `sudo ldconfig` is now useless since we don't use the installed binaries.

We also run `sudo ldconfig` in the command before this one (line 32) and it useless as well. But maybe we should consider running tests on both a) uninstalled binaries and b) installed binaries (in order to make sure we install them at the proper place).

PR #89, opened 2015-04-11
Install bash completions in /usr/share

krocard:
With a classic cmake install prefix '/usr', bash completions were installed in _/usr/etc/bash_completion.d/_ which even before bash-completion 1.9 was incorrect.

Since v1.9 all completions are to be installed in _/usr/share/bash-completion/completions_
See: http://anonscm.debian.org/cgit/bash-completion/bash-completion.git/tree/CHANGES#n324

Change the completions install path accordingly


krocard:

clero:
:+1:

PR #95, opened 2015-04-14
python: prefer 2.7.x and make sure interpreter and libs are aligned

dawagner:
CMake's find_package command makes it possible to specify a version.  Take
advantage of that to favor the interpreter's version to 2.7 and then pin the
library's version to the very same version as the interpreter.

This does not ensure that 2.7.x is used (e.g. if 2.7.x is not present, another
version will be used) but it helps a lot.


dawagner:
@krocard @clero please review

krocard:
I did :+1: informally before the merge.

PR #93, opened 2015-04-14
Manager is not element

clero:
Avoid ParameterMgr inheritance to Element
Allows to avoid tree structure guess work

PR #96, opened 2015-04-15
Criterion rework

clero:
Depends on #93 

This is a rework of Pfw criterion. Please do not merge now but some feedback on designs changes would be appreciate.

Next step:
Update coding style everywhere
Some documentation is still missing

### DIFF
--- a/bindings/c/ParameterFramework.cpp
+++ b/bindings/c/ParameterFramework.cpp
@@ -32,13 +32,13 @@
 #include <ParameterMgrPlatformConnector.h>
 
 #include <iostream>
-#include <limits>
 #include <string>
 #include <map>
 
 #include <cassert>
 #include <cstring>
 #include <cstdlib>
+#include <climits>
 
 using std::string;
 using core::criterion::CriterionInterface;
@@ -193,16 +193,13 @@ bool PfwHandler::createCriteria(const PfwCriterion criteriaArray[], size_t crite
                 pfw->createExclusiveCriterion(criterion.name));
         assert(newCriterion != NULL);
         // Add criterion values
+        int inclusiveCriterionMaxValue = sizeof(int) * CHAR_BIT - 1;
         for (size_t valueIndex = 0; criterion.values[valueIndex] != NULL; ++valueIndex) {
-            int value;
-            if (criterion.inclusive) {
-                // Check that (int)1 << valueIndex would not overflow (UB)
-                if(std::numeric_limits<int>::max() >> valueIndex == 0) {
-                    return status.failure("Too many values for criterion " +
-                                          string(criterion.name));
-                }
-                value = 1 << valueIndex;
-            } else {
+            int value = valueIndex + 1;
+            if (criterion.inclusive && value > inclusiveCriterionMaxValue) {
+                return status.failure("Too many values for criterion " +
+                                      string(criterion.name));
+            } else if (!criterion.inclusive) {
                 value = valueIndex;
             }
             const char * valueName = criterion.values[valueIndex];
--- a/bindings/c/ParameterFramework.cpp
+++ b/bindings/c/ParameterFramework.cpp
@@ -38,7 +38,6 @@
 #include <cassert>
 #include <cstring>
 #include <cstdlib>
-#include <climits>
 
 using std::string;
 using core::criterion::CriterionInterface;
@@ -264,7 +263,7 @@ static CriterionInterface *getCriterion(const pfw::Criteria &criteria, const str
     return it == criteria.end() ? NULL : it->second;
 }
 
-bool pfwSetCriterion(PfwHandler *handle, const char name[], int value)
+bool pfwSetCriterion(PfwHandler *handle, const char name[], int value[], size_t nbValue)
 {
     if (handle == NULL) { return Status::failure(); }
     Status &status = handle->lastStatus;
@@ -280,10 +279,10 @@ bool pfwSetCriterion(PfwHandler *handle, const char name[], int value)
     if (criterion == NULL) {
         return status.failure("Can not set criterion " + string(name) + " as does not exist");
     }
-    criterion->setCriterionState(value);
-    return status.success();
+    core::criterion::State state{value, value + nbValue};
+    return status.forward(criterion->setState(state, status.msg()));
 }
-bool pfwGetCriterion(const PfwHandler *handle, const char name[], int *value)
+bool pfwGetCriterion(const PfwHandler *handle, const char name[], int *value[], size_t *nbValue)
 {
     if (handle == NULL) { return Status::failure(); }
     Status &status = handle->lastStatus;
@@ -299,11 +298,22 @@ bool pfwGetCriterion(const PfwHandler *handle, const char name[], int *value)
         return status.failure("Can not get criterion \"" + string(name) +
                               "\" as the out value is NULL.");
     }
+    if (nbValue == NULL) {
+        return status.failure("Can not get criterion \"" + string(name) +
+                              "\" as the out value size is NULL.");
+    }
     CriterionInterface *criterion = getCriterion(handle->criteria, name);
     if (criterion == NULL) {
         return status.failure("Can not get criterion " + string(name) + " as it does not exist");
     }
-    *value = criterion->getCriterionState();
+    core::criterion::State state = criterion->getState();
+    *nbValue = state.size();
+    *value = static_cast<int*>(std::malloc(*nbValue * sizeof(int)));
+    if (*value == nullptr && *nbValue > 0) {
+        return status.failure("Can not get criterion " + string(name) +
+                              " as there is no memory left");
+    }
+    std::copy(state.begin(), state.end(), *value);
     return status.success();
 }
 
--- a/bindings/c/ParameterFramework.cpp
+++ b/bindings/c/ParameterFramework.cpp
@@ -189,20 +189,7 @@ bool PfwHandler::createCriteria(const PfwCriterion criteriaArray[], size_t crite
         // Create criterion values
         core::criterion::Values values;
         for (size_t valueIndex = 0; criterion.values[valueIndex] != NULL; ++valueIndex) {
-
-            int value;
-            if (criterion.inclusive) {
-                value = valueIndex + 1;
-            } else {
-                value = valueIndex;
-            }
-            const char * valueName = criterion.values[valueIndex];
-            if (values.count(valueName) != 0) {
-                return status.failure("Could not add value " + string(valueName) +
-                                      " to criterion " + criterion.name +
-                                      ": The value is duplicated");
-            }
-            values[valueName] = value;
+            values.emplace_back(criterion.values[valueIndex]);
         }
         // Create criterion
         string error;
@@ -263,7 +250,7 @@ static CriterionInterface *getCriterion(const pfw::Criteria &criteria, const str
     return it == criteria.end() ? NULL : it->second;
 }
 
-bool pfwSetCriterion(PfwHandler *handle, const char name[], int value[], size_t nbValue)
+bool pfwSetCriterion(PfwHandler *handle, const char name[], const char **values)
 {
     if (handle == NULL) { return Status::failure(); }
     Status &status = handle->lastStatus;
@@ -279,42 +266,60 @@ bool pfwSetCriterion(PfwHandler *handle, const char name[], int value[], size_t 
     if (criterion == NULL) {
         return status.failure("Can not set criterion " + string(name) + " as does not exist");
     }
-    core::criterion::State state{value, value + nbValue};
+    core::criterion::State state{};
+    for (size_t valueIndex = 0; values[valueIndex] != NULL; ++valueIndex) {
+        state.emplace(values[valueIndex]);
+    }
     return status.forward(criterion->setState(state, status.msg()));
 }
-bool pfwGetCriterion(const PfwHandler *handle, const char name[], int *value[], size_t *nbValue)
+const char **pfwGetCriterion(const PfwHandler *handle, const char name[])
 {
-    if (handle == NULL) { return Status::failure(); }
+    if (handle == NULL) { return NULL; }
     Status &status = handle->lastStatus;
     if (name == NULL) {
-        return status.failure("char *name of the criterion is NULL, "
-                              "while getting a criterion.");
+        status.failure("char *name of the criterion is NULL, while getting a criterion.");
+        return NULL;
     }
     if (handle->pfw == NULL) {
-        return status.failure("Can not get criterion \"" + string(name) +
-                              "\" as the parameter framework is not started.");
-    }
-    if (value == NULL) {
-        return status.failure("Can not get criterion \"" + string(name) +
-                              "\" as the out value is NULL.");
-    }
-    if (nbValue == NULL) {
-        return status.failure("Can not get criterion \"" + string(name) +
-                              "\" as the out value size is NULL.");
+        status.failure("Can not get criterion \"" + string(name) +
+                       "\" as the parameter framework is not started.");
+        return NULL;
     }
     CriterionInterface *criterion = getCriterion(handle->criteria, name);
     if (criterion == NULL) {
-        return status.failure("Can not get criterion " + string(name) + " as it does not exist");
+        status.failure("Can not get criterion " + string(name) + " as it does not exist");
+        return NULL;
     }
     core::criterion::State state = criterion->getState();
-    *nbValue = state.size();
-    *value = static_cast<int*>(std::malloc(*nbValue * sizeof(int)));
-    if (*value == nullptr && *nbValue > 0) {
-        return status.failure("Can not get criterion " + string(name) +
-                              " as there is no memory left");
+
+    // Allocating one slot for each criterion value plus one for the NULL end delimiter
+    const char **values = static_cast<const char**>(malloc((state.size() + 1) * sizeof(char*)));
+    if (values == nullptr) {
+        status.failure("Can not get criterion " + string(name) + " as there is no memory left");
+        return NULL;
     }
-    std::copy(state.begin(), state.end(), *value);
-    return status.success();
+
+    size_t i = 0;
+    for (auto &critValue : state) {
+        values[i] = strdup(critValue.c_str());
+        if (values[i] == nullptr) {
+            status.failure("Can not get criterion " + string(name) +
+                           " as there is no memory left (errno: " + strerror(errno) + ")");
+            return NULL;
+        }
+        ++i;
+    }
+    values[i] = NULL;
+    return values;
+}
+
+void pfwCriterionValueFree(const char **value)
+{
+    const char **valueSav = value;
+    while(*value != NULL) {
+        std::free(*const_cast<char**>(value++));
+    }
+    std::free(valueSav);
 }
 
 bool pfwApplyConfigurations(const PfwHandler *handle)
--- a/bindings/c/ParameterFramework.cpp
+++ b/bindings/c/ParameterFramework.cpp
@@ -40,12 +40,12 @@
 #include <cstdlib>
 
 using std::string;
-using core::criterion::CriterionInterface;
+using core::criterion::Criterion;
 
 /** Rename long pfw types to short ones in pfw namespace. */
 namespace pfw
 {
-    typedef std::map<string, CriterionInterface *> Criteria;
+    typedef std::map<string, Criterion*> Criteria;
     typedef CParameterMgrPlatformConnector Pfw;
 }
 
@@ -193,7 +193,7 @@ bool PfwHandler::createCriteria(const PfwCriterion criteriaArray[], size_t crite
         }
         // Create criterion
         string error;
-        CriterionInterface *newCriterion = (criterion.inclusive ?
+        Criterion *newCriterion = (criterion.inclusive ?
                 pfw->createInclusiveCriterion(criterion.name, values, error) :
                 pfw->createExclusiveCriterion(criterion.name, values, error));
         if (newCriterion == nullptr) {
@@ -244,7 +244,7 @@ const char *pfwGetLastError(const PfwHandler *handle)
     return handle == NULL ? NULL : handle->lastStatus.msg().c_str();
 }
 
-static CriterionInterface *getCriterion(const pfw::Criteria &criteria, const string &name)
+static Criterion *getCriterion(const pfw::Criteria &criteria, const string &name)
 {
     pfw::Criteria::const_iterator it = criteria.find(name);
     return it == criteria.end() ? NULL : it->second;
@@ -262,7 +262,7 @@ bool pfwSetCriterion(PfwHandler *handle, const char name[], const char **values)
         return status.failure("Can not set criterion \"" + string(name) +
                               "\" as the parameter framework is not started.");
     }
-    CriterionInterface *criterion = getCriterion(handle->criteria, name);
+    Criterion *criterion = getCriterion(handle->criteria, name);
     if (criterion == NULL) {
         return status.failure("Can not set criterion " + string(name) + " as does not exist");
     }
@@ -285,7 +285,7 @@ const char **pfwGetCriterion(const PfwHandler *handle, const char name[])
                        "\" as the parameter framework is not started.");
         return NULL;
     }
-    CriterionInterface *criterion = getCriterion(handle->criteria, name);
+    Criterion *criterion = getCriterion(handle->criteria, name);
     if (criterion == NULL) {
         status.failure("Can not get criterion " + string(name) + " as it does not exist");
         return NULL;
--- a/bindings/c/ParameterFramework.cpp
+++ b/bindings/c/ParameterFramework.cpp
@@ -187,27 +187,32 @@ bool PfwHandler::createCriteria(const PfwCriterion criteriaArray[], size_t crite
                                   "\" already exist");
         }
 
-        // Create criterion
-        CriterionInterface *newCriterion = (criterion.inclusive ?
-                pfw->createInclusiveCriterion(criterion.name) :
-                pfw->createExclusiveCriterion(criterion.name));
-        assert(newCriterion != NULL);
-        // Add criterion values
-        int inclusiveCriterionMaxValue = sizeof(int) * CHAR_BIT - 1;
+        // Create criterion values
+        core::criterion::Values values;
         for (size_t valueIndex = 0; criterion.values[valueIndex] != NULL; ++valueIndex) {
-            int value = valueIndex + 1;
-            if (criterion.inclusive && value > inclusiveCriterionMaxValue) {
-                return status.failure("Too many values for criterion " +
-                                      string(criterion.name));
-            } else if (!criterion.inclusive) {
+
+            int value;
+            if (criterion.inclusive) {
+                value = valueIndex + 1;
+            } else {
                 value = valueIndex;
             }
             const char * valueName = criterion.values[valueIndex];
-            string error;
-            if(not newCriterion->addValuePair(value, valueName, error)) {
+            if (values.count(valueName) != 0) {
                 return status.failure("Could not add value " + string(valueName) +
-                                      " to criterion " + criterion.name + ": " + error);
+                                      " to criterion " + criterion.name +
+                                      ": The value is duplicated");
             }
+            values[valueName] = value;
+        }
+        // Create criterion
+        string error;
+        CriterionInterface *newCriterion = (criterion.inclusive ?
+                pfw->createInclusiveCriterion(criterion.name, values, error) :
+                pfw->createExclusiveCriterion(criterion.name, values, error));
+        if (newCriterion == nullptr) {
+            return status.failure("Could not create criterion '" + string(criterion.name) +
+                                  "': " + error);
         }
         // Add new criterion to criteria list
         criteria[criterion.name] = newCriterion;
--- a/bindings/c/ParameterFramework.h
+++ b/bindings/c/ParameterFramework.h
@@ -156,25 +156,32 @@ const char *pfwGetLastError(const PfwHandler *handle) NONNULL;
 /** Set a criterion value given its name and value.
   * @param handle[in] @see PfwHandler
   * @param name[in] The name of the criterion that need to be changed.
-  * @param value If the criterion is exclusive, the index of the new value.
-  *              If the criterion is inclusive, a bit field where each bit
-  *              correspond to the value index.
+  * @param value If the criterion is exclusive, an array containing the index of the new value.
+  *              If the criterion is inclusive, an array containing all index of new values.
   * For an inclusive criterion defined as such: { "Red", "Green", "Blue", NULL }
-  * to set the value Green and Blue, value has to be 1<<1 | 1<<2 = 0b110 = 6.
+  * to set the value Green and Blue, value has to be [1, 2].
   * For an exclusive criterion defined as such: { "Car", "Boat", "Plane", NULL }
-  * to set the value Plane, value has to be 2.
+  * to set the value Plane, value has to be [2].
+  * @param nbValue[in] The size of value array.
   *
   * Criterion change do not have impact on the parameters value
   * (no configuration applied) until the changes are committed using pfwApplyConfigurations.
   *
   * @return true on success and false on failure.
   */
-bool pfwSetCriterion(PfwHandler *handle, const char name[], int value) NONNULL USERESULT;
+bool pfwSetCriterion(PfwHandler *handle,
+                     const char name[],
+                     int value[],
+                     size_t nbValue) NONNULL USERESULT;
 /** Get a criterion value given its name.
-  * Same usage as pfwSetCriterion except that value is an out param.
+  * Same usage as pfwSetCriterion except that value and nbValue are out parameters.
   * Get criterion will return the last value setted with pfwSetCriterion independantly of pfwCommitCritenio.
+  * The callee MUST free the returned value array using pfwFree after use.
   */
-bool pfwGetCriterion(const PfwHandler *handle, const char name[], int *value) NONNULL USERESULT;
+bool pfwGetCriterion(const PfwHandler *handle,
+                     const char name[],
+                     int *value[],
+                     size_t *nbValue) NONNULL USERESULT;
 
 /** Commit criteria change and change parameters according to the configurations.
   * Criterion do not have impact on the parameters value when changed,
--- a/bindings/c/ParameterFramework.h
+++ b/bindings/c/ParameterFramework.h
@@ -156,32 +156,36 @@ const char *pfwGetLastError(const PfwHandler *handle) NONNULL;
 /** Set a criterion value given its name and value.
   * @param handle[in] @see PfwHandler
   * @param name[in] The name of the criterion that need to be changed.
-  * @param value If the criterion is exclusive, an array containing the index of the new value.
-  *              If the criterion is inclusive, an array containing all index of new values.
+  * @param value If the criterion is exclusive, an array containing the new value.
+  *              If the criterion is inclusive, an array containing all new values.
   * For an inclusive criterion defined as such: { "Red", "Green", "Blue", NULL }
-  * to set the value Green and Blue, value has to be [1, 2].
+  * to set the value Green and Blue, value has to be ["Green", "Blue", NULL].
   * For an exclusive criterion defined as such: { "Car", "Boat", "Plane", NULL }
-  * to set the value Plane, value has to be [2].
-  * @param nbValue[in] The size of value array.
+  * to set the value Plane, value has to be ["Plane", NULL].
   *
   * Criterion change do not have impact on the parameters value
   * (no configuration applied) until the changes are committed using pfwApplyConfigurations.
   *
   * @return true on success and false on failure.
   */
-bool pfwSetCriterion(PfwHandler *handle,
-                     const char name[],
-                     int value[],
-                     size_t nbValue) NONNULL USERESULT;
+bool pfwSetCriterion(PfwHandler *handle, const char name[], const char **values) NONNULL USERESULT;
 /** Get a criterion value given its name.
-  * Same usage as pfwSetCriterion except that value and nbValue are out parameters.
+  * @param handle[in] @see PfwHandler
+  * @param name[in] The name of the criterion that need to be changed.
+  * @return an NULL terminated array of char* which store criterion values if success
+  *         NULL otherwise
   * Get criterion will return the last value setted with pfwSetCriterion independantly of pfwCommitCritenio.
-  * The callee MUST free the returned value array using pfwFree after use.
+  * The callee MUST free the returned value array using pfwCriterionValueFree after use.
   */
-bool pfwGetCriterion(const PfwHandler *handle,
-                     const char name[],
-                     int *value[],
-                     size_t *nbValue) NONNULL USERESULT;
+const char **pfwGetCriterion(const PfwHandler *handle, const char name[]) NONNULL USERESULT;
+
+/** Frees the memory space of criterion value array,
+  * which must have been returned by a previous call to pfwGetCriterion.
+  *
+  * @param value[in] pointer to the memory to free.
+  * @see man 3 free for usage.
+  */
+void pfwCriterionValueFree(const char **value);
 
 /** Commit criteria change and change parameters according to the configurations.
   * Criterion do not have impact on the parameters value when changed,
--- a/bindings/c/Test.cpp
+++ b/bindings/c/Test.cpp
@@ -306,13 +306,17 @@ TEST_CASE_METHOD(Test, "Parameter-framework c api use") {
                 REQUIRE_FAILURE(pfwGetCriterion(pfw, "Do not exist", &criterionGetValue, &nbValue));
                 REQUIRE(nbValue == 0);
             }
-            THEN("All criterion should value 0") {
+            THEN("All criterion should have their default value") {
                 for(size_t i = 0; i < criterionNb; ++i) {
                     const char *criterionName = criteria[i].name;
                     CAPTURE(criterionName);
                     REQUIRE_SUCCESS(pfwGetCriterion(pfw, criterionName, &criterionGetValue, &nbValue));
-                    REQUIRE(nbValue == 1);
-                    REQUIRE(*criterionGetValue == 0);
+                    if (criteria[i].inclusive) {
+                        REQUIRE(nbValue == 0);
+                    } else {
+                        REQUIRE(nbValue == 1);
+                        REQUIRE(*criterionGetValue == 0);
+                    }
                     pfwFree((void*) criterionGetValue);
                 }
             }
--- a/bindings/c/Test.cpp
+++ b/bindings/c/Test.cpp
@@ -204,8 +204,8 @@ TEST_CASE_METHOD(Test, "Parameter-framework c api use") {
         }
         GIVEN("A criteria with lots of values")
         {
-            // Build a criterion with as many value as there is bits in int.
-            std::vector<char> names(sizeof(int) * CHAR_BIT + 1, 'a');
+            // Build a criterion with a lot of value
+            std::vector<char> names(300, 'a');
             names.back() = '\0';
             std::vector<const char *> values(names.size());
             for(size_t i = 0; i < values.size(); ++i) {
@@ -240,11 +240,7 @@ TEST_CASE_METHOD(Test, "Parameter-framework c api use") {
              */
             const PfwCriterion duplicatedCriteria[] = {{"name", true, &values[0]}};
 
-            WHEN("The pfw is started with a too long criterion state list") {
-                REQUIRE_FAILURE(pfwStart(pfw, config, duplicatedCriteria, 1, &logger));
-            }
-            WHEN("The pfw is started with max length criterion state list") {
-                values[values.size() - 2] = NULL; // Hide last value
+            WHEN("The pfw is started with a long criterion state list") {
                 REQUIRE_SUCCESS(pfwStart(pfw, config, duplicatedCriteria, 1, &logger));
             }
         }
@@ -267,11 +263,14 @@ TEST_CASE_METHOD(Test, "Parameter-framework c api use") {
         }
 
         WHEN("Get criterion of a stopped pfw") {
-            int value;
-            REQUIRE_FAILURE(pfwGetCriterion(pfw, criteria[0].name, &value));
+            int *value = nullptr;
+            size_t nbValue = 0;
+            REQUIRE_FAILURE(pfwGetCriterion(pfw, criteria[0].name, &value, &nbValue));
+            REQUIRE(nbValue == 0);
         }
         WHEN("Set criterion of a stopped pfw") {
-            REQUIRE_FAILURE(pfwSetCriterion(pfw, criteria[0].name, 1));
+            std::vector<int> value{1};
+            REQUIRE_FAILURE(pfwSetCriterion(pfw, criteria[0].name, value.data(), value.size()));
         }
         WHEN("Commit criteria of a stopped pfw") {
             REQUIRE_FAILURE(pfwApplyConfigurations(pfw));
@@ -284,55 +283,70 @@ TEST_CASE_METHOD(Test, "Parameter-framework c api use") {
         WHEN("The pfw is started correctly")
         {
             REQUIRE_SUCCESS(pfwStart(pfw, config, criteria, criterionNb, &logger));
-            int value;
+            int value = 0;
+            int *criterionGetValue = nullptr;
+            size_t nbValue = 0;
 
             WHEN("Get criterion without an handle") {
-                REQUIRE(not pfwGetCriterion(NULL, criteria[0].name, &value));
+                REQUIRE(not pfwGetCriterion(NULL, criteria[0].name, &criterionGetValue, &nbValue));
+                REQUIRE(nbValue == 0);
             }
             WHEN("Get criterion without a name") {
-                REQUIRE_FAILURE(pfwGetCriterion(pfw, NULL, &value));
+                REQUIRE_FAILURE(pfwGetCriterion(pfw, NULL, &criterionGetValue, &nbValue));
+                REQUIRE(nbValue == 0);
             }
             WHEN("Get criterion without an output value") {
-                REQUIRE_FAILURE(pfwGetCriterion(pfw, criteria[0].name, NULL));
+                REQUIRE_FAILURE(pfwGetCriterion(pfw, criteria[0].name, NULL, &nbValue));
+                REQUIRE(nbValue == 0);
+            }
+            WHEN("Get criterion without an output value size") {
+                REQUIRE_FAILURE(pfwGetCriterion(pfw, criteria[0].name, &criterionGetValue, NULL));
             }
             WHEN("Get not existing criterion") {
-                REQUIRE_FAILURE(pfwGetCriterion(pfw, "Do not exist", &value));
+                REQUIRE_FAILURE(pfwGetCriterion(pfw, "Do not exist", &criterionGetValue, &nbValue));
+                REQUIRE(nbValue == 0);
             }
             THEN("All criterion should value 0") {
                 for(size_t i = 0; i < criterionNb; ++i) {
                     const char *criterionName = criteria[i].name;
                     CAPTURE(criterionName);
-                    REQUIRE_SUCCESS(pfwGetCriterion(pfw, criterionName, &value));
-                    REQUIRE(value == 0);
+                    REQUIRE_SUCCESS(pfwGetCriterion(pfw, criterionName, &criterionGetValue, &nbValue));
+                    REQUIRE(nbValue == 1);
+                    REQUIRE(*criterionGetValue == 0);
+                    pfwFree((void*) criterionGetValue);
                 }
             }
 
             WHEN("Set criterion without an handle") {
-                REQUIRE(not pfwSetCriterion(NULL, criteria[0].name, 1));
+                std::vector<int> setValue{1};
+                REQUIRE(not pfwSetCriterion(NULL, criteria[0].name, setValue.data(), setValue.size()));
             }
             WHEN("Set criterion without a name") {
-                REQUIRE_FAILURE(pfwSetCriterion(pfw, NULL, 2));
+                std::vector<int> setValue{2};
+                REQUIRE_FAILURE(pfwSetCriterion(pfw, NULL, setValue.data(), setValue.size()));
             }
             WHEN("Set not existing criterion") {
-                REQUIRE_FAILURE(pfwSetCriterion(pfw, "Do not exist", 3));
+                std::vector<int> setValue{3};
+                REQUIRE_FAILURE(pfwSetCriterion(pfw, "Do not exist", setValue.data(), setValue.size()));
             }
             WHEN("Set criterion value") {
+                std::vector<int> setValue{2};
                 for(size_t i = 0; i < criterionNb; ++i) {
                     const char *criterionName = criteria[i].name;
                     CAPTURE(criterionName);
-                    REQUIRE_SUCCESS(pfwSetCriterion(pfw, criterionName, 3));
-                }
-                THEN("Get criterion value should return what was set") {
-                    for(size_t i = 0; i < criterionNb; ++i) {
-                        const char *criterionName = criteria[i].name;
-                        CAPTURE(criterionName);
-                        REQUIRE_SUCCESS(pfwGetCriterion(pfw, criterionName, &value));
-                        REQUIRE(value == 3);
+                    REQUIRE_SUCCESS(pfwSetCriterion(pfw, criterionName, setValue.data(), setValue.size()));
+                    THEN("Get criterion value should return what was set") {
+                        REQUIRE_SUCCESS(pfwGetCriterion(pfw, criterionName, &criterionGetValue, &nbValue));
+                        REQUIRE(nbValue == 1);
+                        REQUIRE(*criterionGetValue == 2);
+                        pfwFree((void*) criterionGetValue);
                     }
                 }
                 WHEN("Set a new value to a criterion without committing first") {
+                    setValue[0] = 1;
                     const char *criterionName = criteria[0].name;
-                    REQUIRE_SUCCESS(pfwSetCriterion(pfw, criterionName, 0));
+                    REQUIRE_SUCCESS(pfwSetCriterion(pfw, criterionName,
+                                                    setValue.data(), setValue.size()));
                     THEN("A warning message should have been displayed") {
                         INFO("Previous pfw log: \n" + logLines);
                         size_t logPos = logLines.find("Warning: Selection criterion "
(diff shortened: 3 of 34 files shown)


parameter/criterion/include/criterion/client/CriterionInterface.h
```diff
@@ -0,0 +1,103 @@
...
+namespace criterion
+{
+
+class CriterionInterface
+{
+public:
+    /** Type which associate literal and numerical value */
+    using Values = std::map<std::string, int>;
+    /** */
+    using State = std::set<int>;
```
clero: use an alias for int

parameter/criterion/include/criterion/client/CriterionInterface.h
```diff
@@ -0,0 +1,103 @@
...
+namespace selection
+{
+namespace criterion
+{
+
+class CriterionInterface
+{
+public:
+    /** Type which associate literal and numerical value */
+    using Values = std::map<std::string, int>;
```
clero: Use only string for value ?

parameter/criterion/include/criterion/client/CriterionInterface.h
```diff
@@ -0,0 +1,103 @@
...
+     *
+     */
+    virtual bool setState(const State& state, std::string& error) = 0;
+
+    /**
+     *
+     */
+    virtual State getState() const = 0;
+
+    virtual std::string getCriterionName() const = 0;
```
clero: rename getName

parameter/criterion/src/Criterion.cpp
```diff
@@ -0,0 +1,296 @@
...
+
+std::string Criterion::getCriterionName() const
+{
+    return mName;
+}
+
+std::string Criterion::getFormattedDescription(bool bWithTypeInfo, bool bHumanReadable) const
+{
+    std::string strFormattedDescription;
+    std::string typeName = isInclusive() ? "Inclusive" : "Exclusive";
```
clero: replace by a private getKind method

parameter/criterion/include/criterion/client/CriterionInterface.h
```diff
@@ -0,0 +1,103 @@
...
+     *
+     * @return formatted string of criterion state
+     */
+    virtual std::string getFormattedState() const = 0;
+
+    /** Retrieve Criterion type
+     *
+     * @return true if the criterion is Inclusive, false if it is Exclusive
+     */
+    virtual bool isInclusive() const = 0;
```
clero: Is this really necessary ?

clero:
@dawagner @krocard @Makohoek @OznOg please review

bindings/c/ParameterFramework.cpp
```diff
@@ -32,20 +32,21 @@
...
+#include <climits>
 
 using std::string;
+using core::selection::criterion::CriterionInterface;
 
 /** Rename long pfw types to short ones in pfw namespace. */
 namespace pfw
 {
-    typedef ISelectionCriterionInterface Criterion;
+    typedef CriterionInterface Criterion;
```
dawagner: You are not using this typedef.

parameter/CMakeLists.txt
```diff
@@ -107,11 +104,13 @@ include_directories(
...
     "${PROJECT_SOURCE_DIR}/remote-processor"
-    "${PROJECT_SOURCE_DIR}/parameter/log/include")
+    "${PROJECT_SOURCE_DIR}/parameter/"
+    "${PROJECT_SOURCE_DIR}/parameter/log/include"
+    "${PROJECT_SOURCE_DIR}/parameter/criterion/include")
 
 # No need to link with libremote-processor: it is accessed via dlopen()
 find_library(dl dl)
-target_link_libraries(parameter xmlserializer pfw_utility dl)
+target_link_libraries(parameter xmlserializer pfw_utility criterion)
```
dawagner: You probably didn't mean to removing the linking to libdl. see 7f7a206ed4f765e80c24fa88ddfea24a4bbb9442

parameter/CMakeLists.txt
```diff
@@ -107,11 +104,13 @@ include_directories(
     "${PROJECT_SOURCE_DIR}/xmlserializer"
     "${PROJECT_SOURCE_DIR}/utility"
     "${PROJECT_SOURCE_DIR}/remote-processor"
-    "${PROJECT_SOURCE_DIR}/parameter/log/include")
+    "${PROJECT_SOURCE_DIR}/parameter/"
```
dawagner: not needed (implicit).

parameter/CompoundRule.cpp
```diff
@@ -34,6 +34,8 @@
 
 using std::string;
 
+using namespace core::selection;
+
```
dawagner: seems unused

parameter/ConfigurableDomain.cpp
```diff
@@ -769,7 +769,10 @@ bool CConfigurableDomain::getElementSequence(const string& strConfiguration, str
     return true;
 }
 
-bool CConfigurableDomain::setApplicationRule(const string& strConfiguration, const string& strApplicationRule, const CSelectionCriteriaDefinition* pSelectionCriteriaDefinition, string& strError)
+bool CConfigurableDomain::setApplicationRule(const string& strConfiguration,
+                                             const string& strApplicationRule,
+                                             const core::selection::criterion::Criteria& criteria,
```
dawagner: The 'selection' namespace doesn't seem very useful yet: everything it contains is contained is its sub-namespace 'criterion'.

parameter/ConfigurableDomains.h
```diff
@@ -31,6 +31,8 @@
 
 #include "BinarySerializableElement.h"
 #include "Results.h"
+#include <criterion/Criteria.h>
```
dawagner: please use quotes.

parameter/DomainConfiguration.h
```diff
@@ -31,6 +31,8 @@
 
 #include "BinarySerializableElement.h"
 #include "Results.h"
+#include <criterion/Criteria.h>
```
dawagner: ditto

parameter/DomainConfiguration.h
```diff
@@ -60,8 +61,17 @@ class CDomainConfiguration : public CBinarySerializableElement
     bool setElementSequence(const std::vector<std::string>& astrNewElementSequence, std::string& strError);
     void getElementSequence(std::string& strResult) const;
 
-    // Application rule
-    bool setApplicationRule(const std::string& strApplicationRule, const CSelectionCriteriaDefinition* pSelectionCriteriaDefinition, std::string& strError);
+    /** Create a new application rule for this configuration
```
dawagner: "Create a new" -> "Set the"

parameter/ParameterMgr.h
```diff
@@ -34,16 +34,15 @@
...
 #include "Element.h"
 #include "XmlDocSink.h"
 #include "XmlDocSource.h"
 #include "Results.h"
 #include "ParameterFrameworkConfiguration.h"
-#include "SelectionCriteria.h"
+#include "criterion/Criteria.h"
 #include "ConfigurableDomains.h"
 #include "SystemClass.h"
+#include <criterion/Criterion.h>
```
dawagner: please use quotes

parameter/ParameterMgr.h
```diff
@@ -104,11 +103,28 @@ class CParameterMgr
       */
     bool load(std::string& strError);
 
-    // Selection Criteria
-    CSelectionCriterionType* createSelectionCriterionType(bool bIsInclusive);
-    CSelectionCriterion* createSelectionCriterion(const std::string& strName, const CSelectionCriterionType* pSelectionCriterionType);
+    /** Create a new Exclusive criterion
+     *
+     * @param[in] name, the criterion name
```
dawagner: missing other param's doc

parameter/ParameterMgr.h
```diff
@@ -104,11 +103,28 @@ class CParameterMgr
...
+     * @return raw pointer on the created criterion
+     */
+    core::selection::criterion::Criterion*
+    createExclusiveCriterion(const std::string& name,
+                             const core::selection::criterion::CriterionInterface::Values& values,
+                             std::string& error);
+
+    /** Create a new Inclusive criterion
+     *
+     * @param[in] name, the criterion name
```
dawagner: ditto

parameter/ParameterMgr.h
```diff
@@ -104,11 +103,28 @@ class CParameterMgr
...
 
-    // Selection Criteria
-    CSelectionCriterionType* createSelectionCriterionType(bool bIsInclusive);
-    CSelectionCriterion* createSelectionCriterion(const std::string& strName, const CSelectionCriterionType* pSelectionCriterionType);
+    /** Create a new Exclusive criterion
+     *
+     * @param[in] name, the criterion name
+     * @return raw pointer on the created criterion
+     */
+    core::selection::criterion::Criterion*
```
dawagner: Are you sure the client should receive a Criterion object ? Shouldn't it receive a CriterionInterface object instead.

parameter/ParameterMgr.h
dawagner: Anyway, I don't think the client has access to this header

parameter/ParameterMgrFullConnector.cpp
```diff
@@ -34,6 +34,8 @@
 #include <list>
 
 using std::string;
+using core::selection::criterion::CriterionInterface;
+using core::selection::criterion::Criterion;
```
dawagner: unused typedef + client shouldn't have access to it.

parameter/ParameterMgrPlatformConnector.cpp
```diff
@@ -33,6 +33,8 @@
 #include <assert.h>
 
 using std::string;
+using core::selection::criterion::CriterionInterface;
+using core::selection::criterion::Criterion;
```
dawagner: unused "using"

parameter/criterion/CMakeLists.txt
```diff
@@ -0,0 +1,65 @@
...
+
+target_link_libraries(criterion pfw_utility xmlserializer)
+
+# '-fPIC' needed for linking against shared libraries (e.g. libparameter)
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fPIC")
+
+# Client headers
+install(FILES
+    include/criterion/client/CriterionInterface.h
+    DESTINATION "include/parameter/criterion/client")
```
dawagner: I think it would be more consistent to install under `include/parameter/client/criterion` in order to have all client headers under `include/parameter/client`.

parameter/criterion/CMakeLists.txt
```diff
@@ -0,0 +1,65 @@
...
+
+include_directories(
+    include
+    "${PROJECT_SOURCE_DIR}/parameter/log/include"
+    "${PROJECT_SOURCE_DIR}/xmlserializer/"
+    "${PROJECT_SOURCE_DIR}/utility/")
+
+target_link_libraries(criterion pfw_utility xmlserializer)
+
+# '-fPIC' needed for linking against shared libraries (e.g. libparameter)
```
dawagner: needed for shared library to link against libcriterion

parameter/SelectionCriterionRule.cpp
```diff
@@ -28,26 +28,19 @@
...
-const CSelectionCriterionRule::SMatchingRuleDescription CSelectionCriterionRule::_astMatchesWhen[CSelectionCriterionRule::ENbMatchesWhen] = {
-    { "Is", true },
-    { "IsNot", true },
-    { "Includes", false },
-    { "Excludes", false }
-};
-
-CSelectionCriterionRule::CSelectionCriterionRule() : _pSelectionCriterion(NULL), _eMatchesWhen(CSelectionCriterionRule::EIs), _iMatchValue(0)
+CSelectionCriterionRule::CSelectionCriterionRule() :
+    _pSelectionCriterion(NULL), mMatchesWhenVerb(""), mMatchState{0}
```
dawagner: In this class, mMatchState is always a singleton set by design; maybe you can just handle an integer in this class and (maybe) wrap it in a set only when calling Criterion::match

parameter/SelectionCriterionRule.h
```diff
@@ -30,29 +30,14 @@
 #pragma once
 
 #include "Rule.h"
+#include <criterion/Criterion.h>
 
 #include <string>
 
-class CSelectionCriterion;
+// FIXME: Add SelectionCriterionRule in core::selection namespace
```
dawagner: Is it to be done later or now ?

bindings/python/pfw.i
```diff
@@ -80,10 +80,18 @@ public:
...
+                             const core::selection::criterion::CriterionInterface::Values& values,
+                             std::string& error);
+
+    core::selection::criterion::CriterionInterface*
+    createInclusiveCriterion(const std::string& name,
+                             const core::selection::criterion::CriterionInterface::Values& values,
+                             std::string& error);
+
+    core::selection::criterion::CriterionInterface*
+    getSelectionCriterion(const std::string& name);
```
dawagner: please modify sample.py to take these changes into account. Maybe we should have a test for this.

parameter/criterion/include/criterion/Criteria.h
```diff
@@ -0,0 +1,155 @@
...
+ */
+#pragma once
+
+#include "criterion/Criterion.h"
+#include "XmlSource.h"
+#include <log/Logger.h>
+
+#include <string>
+#include <list>
+#include <memory>
```
dawagner: include <map> too

parameter/criterion/include/criterion/Criteria.h
```diff
@@ -0,0 +1,155 @@
...
+        try {
+            Criterion* criterion(new CriterionType(name, values, logger));
+            mCriteria.emplace(name, CriterionWrapper(criterion));
+            return criterion;
+        }
+        catch (Criterion::InvalidCriterionError& e) {
+            error = e.what();
+            return nullptr;
+        }
+    }
```
dawagner: maybe defined this in the .cpp ?

parameter/criterion/include/criterion/Criteria.h
```diff
@@ -0,0 +1,155 @@
...
+
+    /** Criteria instance container */
+    CriteriaMap mCriteria;
```
dawagner: I suppose this class (Criteria) should be made non-copyable or `unique_ptr` replaced with `shared_ptr` in the definition of CriterionWrapper.

parameter/criterion/include/criterion/Criterion.h
```diff
@@ -0,0 +1,180 @@
...
+public:
+    using InvalidCriterionError = std::runtime_error;
+
+    /** @param[in] name the criterion name
+     *  @param[in] logger the main application logger
+     */
+    Criterion(const std::string& name, const Values& values, core::log::Logger& logger);
+
+    // @{
+    /** @see CriterionInterface */
```
dawagner: If you want this description to be set for the whole group of methods, the correct way is: http://www.stack.nl/~dimitri/doxygen/manual/commands.html#cmdaddtogroup

parameter/criterion/include/criterion/Criterion.h
```diff
@@ -0,0 +1,180 @@
...
+    bool getLiteralValue(int numericalValue, std::string& literalValue) const override final;
+
+    virtual bool getNumericalValue(const std::string& literalValue,
+                                   int& numericalValue) const override final;
+
+    virtual std::string getFormattedState() const override;
+    // @}
+
+    bool hasBeenModified() const;
+    void resetModifiedStatus();
```
dawagner: doc ?

parameter/criterion/include/criterion/Criterion.h
```diff
@@ -0,0 +1,180 @@
...
+    /** Match method container, MatchMethod are indexed by their name */
+    typedef std::map<std::string, MatchMethod> MatchMethods;
+
+    /** Initializer constructor
+     * This Constructor initialize class members and should be called by derived class
+     * in order to add functionalities
+     *
+     * @param[in] name the criterion name
+     * @param[in] logger the main application logger
+     * @param[in] derivedValuePairs initial value pairs of derived classes
```
dawagner: missing defaultState doc

parameter/criterion/include/criterion/Criterion.h
```diff
@@ -0,0 +1,180 @@
...
+              const MatchMethods& derivedMatchMethods);
+
+    /** Contains pair association between literal and numerical value */
+    const Values mValues;
+
+    /** Available criterion match methods */
+    const MatchMethods mMatchMethods;
+
+    /** Current state */
+    State mState;
```
dawagner: member attributes should be private (at the very least non-const ones).

parameter/criterion/src/Criterion.cpp
```diff
@@ -0,0 +1,296 @@
...
+{
+    return mMatchMethods.count(method) == 1;
+}
+
+template<class MapType>
+MapType Criterion::gatherMap(MapType mapA, const MapType& mapB)
+{
+    mapA.insert(mapB.begin(), mapB.end());
+    return mapA;
+}
```
dawagner: Suggestion: have a non-virtual protected "addMatchMethod" method and call it from the derived constructor.

Other suggestion: make the map of methods a private static const value for each criterion type and have a virtual "getMatchMethods()" method on which `Criterion::match` and Criterion::iMatchMethodAvailable` would rely.

One more suggestion: make `Criterion` template over a boolean (or an enum).

parameter/criterion/src/Criterion.cpp
```diff
@@ -0,0 +1,296 @@
...
+
+template
+Criterion::MatchMethods Criterion::gatherMap<Criterion::MatchMethods>(MatchMethods mapA,
+                                                                      const MatchMethods& mapB);
+template
+Criterion::Values Criterion::gatherMap<Criterion::Values>(Values mapA,
+                                                          const Values& mapB);
```
dawagner: seems unused

parameter/criterion/src/Criterion.cpp
```diff
@@ -0,0 +1,296 @@
...
+Criterion::Criterion(const std::string& name,
+                     core::log::Logger& logger,
+                     const Values& values,
+                     const State&& defaultState,
+                     const MatchMethods& derivedMatchMethods)
+    : mValues(values),
+      mMatchMethods(gatherMap<MatchMethods>(MatchMethods{
+                                {"Is", [&](const State& state){ return mState == state; }},
+                                {"IsNot", [&](const State& state){ return mState != state; }}},
+                                derivedMatchMethods)),
```
dawagner: I find this very strange... see comments at the end of this file.

parameter/criterion/src/Criterion.cpp
```diff
@@ -0,0 +1,296 @@
...
+bool Criterion::isInclusive() const
+{
+    return false;
+}
+
+bool Criterion::getNumericalValue(const std::string& literalValue,
+                                  int& numericalValue) const
+{
+    try {
+        numericalValue = mValues.at(literalValue);
```
dawagner: Wouldn't you be better of using find ?

parameter/criterion/src/Criterion.cpp
```diff
@@ -0,0 +1,296 @@
...
+        return true;
+    }
+    catch (std::out_of_range& e) {
+        return false;
+    }
+}
+
+bool Criterion::getLiteralValue(int numericalValue, std::string& literalValue) const
+{
+    for (auto& value : mValues) {
```
dawagner: ditto

parameter/criterion/src/Criterion.cpp
```diff
@@ -0,0 +1,296 @@
...
+    if (state.size() > 1) {
+        error = "Exclusive criterion '" + mName + "' can't be set with more than one value";
+        return false;
+    }
+
+    // Check for a change
+    if (mState != state) {
+        bool stateChanged = false;
+        if (state.empty()) {
+            stateChanged = setDefaultState();
```
dawagner: IMO, setDefaultState should use setState, not the other way around.

parameter/criterion/src/Criterion.cpp
```diff
@@ -0,0 +1,296 @@
...
+{
+    return _uiNbModifications != 0;
+}
+
+void Criterion::resetModifiedStatus()
+{
+    _uiNbModifications = 0;
+}
+
+bool Criterion::setDefaultState()
```
dawagner: see below. IMO, the code for this method should be:

    return setState(mDefaultState);

parameter/criterion/src/Criterion.cpp
dawagner: if (state.empty()) {
        state = mDefaultState;
    }

parameter/criterion/include/criterion/client/CriterionInterface.h
```diff
@@ -0,0 +1,103 @@
...
+class CriterionInterface
+{
+public:
+    /** Type which associate literal and numerical value */
+    using Values = std::map<std::string, int>;
+    /** */
+    using State = std::set<int>;
+
+    /**
+     *
```
dawagner: doc

parameter/criterion/include/criterion/client/CriterionInterface.h
```diff
@@ -0,0 +1,103 @@
...
+     */
+    virtual bool setState(const State& state, std::string& error) = 0;
+
+    /**
+     *
```
dawagner: doc

parameter/criterion/src/InclusiveCriterion.cpp
```diff
@@ -0,0 +1,109 @@
...
+InclusiveCriterion::InclusiveCriterion(const std::string& name,
+                                       const Values& values,
+                                       core::log::Logger& logger)
+    : Criterion(name, logger, values, {},
+                {{"Includes", [&](const State& state) {
+                    return std::includes(mState.begin(), mState.end(),
+                                         state.begin(), state.end()); }},
+                 {"Excludes", [&](const State& state) {
+                    State inter;
+                    std::set_intersection(mState.begin(), mState.end(),
```
dawagner: Since, for now, the candidate state can only be a singleton set, you can return `!std::includes(...)`. You can even simplify further by taking only an int (or did I miss something ?)

test/functional-tests/Util/PfwUnitTestLib.py
```diff
@@ -67,10 +67,10 @@ def stopHal(self):
         subprocess.call("remote-process localhost 5001 exit", shell=True)
 
     def createInclusiveCriterion(self, name, nb):
-        self.sendCmd("createInclusiveSelectionCriterion", name, nb)
+        self.sendCmd("createInclusiveCriterion", name, nb)
 
     def createExclusiveCriterion(self, name, nb):
-        self.sendCmd("createExclusiveSelectionCriterion", name, nb)
+        self.sendCmd("createExclusiveCriterion", name, nb)
```
dawagner: we could consider keeping backward compat for remote-process scripts.

test/test-platform/TestPlatform.cpp
```diff
@@ -289,153 +290,124 @@ CTestPlatform::CommandReturn CTestPlatform::applyConfigurations(const IRemoteCom
...
-                                                      string& strResult)
+bool CTestPlatform::createInclusiveCriterion(const string& name, uint32_t nbStates, string& result)
 {
-    ISelectionCriterionTypeInterface* pCriterionType =
-        _pParameterMgrPlatformConnector->createSelectionCriterionType(true);
+    using namespace core::selection::criterion;
+    CriterionInterface::Values values;
 
-    if (uiNbStates > 32) {
+    if (nbStates > 32) {
```
dawagner: That's not true anymore, is it ?

parameter/CMakeLists.txt
clero: done

parameter/CMakeLists.txt
clero: done

parameter/CompoundRule.cpp
clero: done

parameter/ConfigurableDomain.cpp
clero: you're right, I used it for consistency. 
In fact I thought it could be useful to later add rules to the selection namespace but for now YAGNI.

parameter/ConfigurableDomains.h
clero: As criterion is know a separate library I though it can be good to use <> form. Is it for system headers only?

parameter/ParameterMgr.h
clero: You're right, In fact I mess up as my first intention was to rename CriterionInterface to Criterion.

parameter/ParameterMgrFullConnector.cpp
clero: done

parameter/ParameterMgrPlatformConnector.cpp
clero: done

parameter/criterion/CMakeLists.txt
clero: done

parameter/criterion/CMakeLists.txt
clero: done

parameter/SelectionCriterionRule.h
clero: To be done when rules will be abstracted I think

parameter/criterion/include/criterion/Criterion.h
clero: state is protected because the derived class handle it differently, so maybe we need a third class which will be abstract ?

parameter/criterion/src/Criterion.cpp
clero: non-virtual protected "addMatchMethod" method: seems ok to me but we loose the initialization list use (but does it really matter?)

Suggestion 2: Exclusive criterion match method would be duplicated no ?

clero: In fact no Suggestion 2 seems to be the best thx!

parameter/criterion/src/Criterion.cpp
clero: at does not add a new element if it is not found (like [] do)
So I prefer this notation rather the iterator one we have to use with find

parameter/criterion/src/Criterion.cpp
clero: find only works with "key" map, as we search for the mapped value I think we need to test all keys

parameter/criterion/include/criterion/Criteria.h
clero: done

parameter/SelectionCriterionRule.cpp
clero: It wasn't clear that the state was a singleton (and for me it's too bad).
Anyway, I think it's much clearer to keep the State type everywhere.

parameter/ParameterMgr.h
clero: done

parameter/ParameterMgr.h
clero: done

parameter/criterion/include/criterion/Criteria.h
```diff
@@ -0,0 +1,155 @@
...
+     *
+     * @param[out] results information container
+     * @param[in] withTypeInfo indicates if we want to retrieve criterion type information
+     * @param[in] humanReadable indicates the formatage we want to use
+     */
+    void listSelectionCriteria(std::list<std::string>& results,
+                               bool withTypeInfo,
+                               bool humanReadable) const;
+
+    /** Reset the modified status of criterions */
```
clero: s/criterions/criteria

parameter/criterion/include/criterion/Criterion.h
clero: done

parameter/criterion/src/Criterion.cpp
clero: your proposition is not exactly right
The default state can be set either by setting an empty state or by explicitly setting the state which is the default one.
if we use 
```cpp
if (state.empty()) {
    state = mDefaultState;
}
```
then we will have the modification counter incremented in the wrong way (or not incremented at all).

This is why i wrote the setDefaultState function which is here only for better readability.

parameter/criterion/include/criterion/client/CriterionInterface.h
clero: done

bindings/c/ParameterFramework.cpp
clero: done

parameter/criterion/include/criterion/Criteria.h
clero: done

parameter/criterion/src/InclusiveCriterion.cpp
clero: When you talk about the singleton, are you saying that because Rules only use singleton values?
Here I try to assume nothing about how the criterion will be used.

parameter/criterion/include/criterion/client/CriterionInterface.h
clero: done

parameter/criterion/include/criterion/Criteria.h
clero: done

parameter/criterion/src/Criterion.cpp
clero: done

parameter/criterion/include/criterion/Criterion.h
clero: done

test/test-platform/TestPlatform.cpp
clero: I let to test-platform the exact same bahavior. Nevertheless, other clients can have as many inclusive criterion value as they want.
The problem with test-platform is that the automated criterion creation (we only give a number) if used in tests (functional tests for example) and use a mask to name criteria. So if we change test-platform behavior, all tests must be updated.

bindings/python/pfw.i
clero: I've started this work but I have some difficulties with swig semantics due to nested typedef I think, I need to read some documentation

parameter/criterion/include/criterion/Criteria.h
clero: done

parameter/criterion/src/InclusiveCriterion.cpp
dawagner: I'm saying this because rules can only be

    SomeCriterionName Excludes SomeValue

i.e., it can't be `SomeCriterionName Excludes {SomeValueA, SomeValueB}`.

test/test-platform/TestPlatform.cpp
dawagner: In this case, I think tests must be updated.

bindings/c/ParameterFramework.cpp
```diff
@@ -32,21 +32,21 @@
...
 
 #include <iostream>
-#include <limits>
 #include <string>
 #include <map>
 
 #include <cassert>
 #include <cstring>
 #include <cstdlib>
+#include <climits>
```
krocard: why including `<climits>`?

parameter/criterion/src/Criteria.cpp
```diff
@@ -0,0 +1,129 @@
...
+{
+}
+
+Criterion* Criteria::getCriterionPointer(const std::string& name) const
+{
+    // Bound exception aware code to criteria, others Pfw parts will check nullptr.
+    try {
+        return mCriteria.at(name).get();
+    }
+    catch (std::out_of_range& e) {
```
krocard: no need for the ` e`.

parameter/criterion/include/criterion/Criterion.h
krocard: It is usual in this case to provide accessors.

dawagner:
general remark on commit messages: I'm not fond of the "[component]" tag style. I prefer the "component: " tag style and sometimes, you can avoid repeating the component name in the commit message.

e.g.:

    [Criterion] Add unit test to criterion library

becomes

    libcriterion: add unit tests

parameter/criterion/include/criterion/client/CriterionInterface.h
```diff
@@ -0,0 +1,108 @@
...
+     * @param[out] error the string describing the error if an error occurred
+     *                   undefined otherwise
+     * @return true if in case of success, false otherwise
+     */
+    virtual bool setState(const State& state, std::string& error) = 0;
+
+    /** Retrieve the current criterion state */
+    virtual State getState() const = 0;
+
+    virtual std::string getCriterionName() const = 0;
```
dawagner: "getName" ?

parameter/criterion/src/Criteria.cpp
```diff
@@ -0,0 +1,129 @@
...
+
+Criterion* Criteria::createInclusiveCriterion(const std::string& name,
+                                              const CriterionInterface::Values& values,
+                                              core::log::Logger& logger,
+                                              std::string& error)
+{
+    return addCriterion<InclusiveCriterion>(name, values, logger, error);
+}
+
+Criterion* Criteria::getSelectionCriterion(const std::string& name)
```
dawagner: "getCriterion" ?

parameter/criterion/src/Criteria.cpp
```diff
@@ -0,0 +1,129 @@
...
+{
+    return getCriterionPointer(name);
+}
+
+const Criterion* Criteria::getSelectionCriterion(const std::string& name) const
```
dawagner: ditto

parameter/criterion/src/Criteria.cpp
```diff
@@ -0,0 +1,129 @@
...
+{
+    return getCriterionPointer(name);
+}
+
+void Criteria::listSelectionCriteria(std::list<std::string>& results,
```
dawagner: "listCriteria" ?

parameter/criterion/src/Criterion.cpp
dawagner: That's not an issue: regardless of what value we are setting, we can check whether it is the same as the current one.

    State oldState = mState;
    if (state.empty()) {
        state = mDefaultState;
    } else {
        if (!getLiteralValue(*state.begin(), error)) {
            error = ...;
            return false;
        }
    }

    mState = state;

    if (mState != oldState) {
        stateModificationsEvent();
    }

    return true;

dawagner: btw, the `if (mState != state)` check at the top becomes useless.

parameter/criterion/src/Criterion.cpp
```diff
@@ -0,0 +1,292 @@
...
+}
+
+bool Criterion::getNumericalValue(const std::string& literalValue,
+                                  int& numericalValue) const
+{
+    try {
+        numericalValue = mValues.at(literalValue);
+        return true;
+    }
+    catch (std::out_of_range& e) {
```
krocard: remove training ` e`.
:package: Remove Strategy Design pattern use for Criterion type

parameter/include/ParameterMgrFullConnector.h
```diff
@@ -71,10 +70,38 @@ class CParameterMgrFullConnector
...
+     *                   undefined otherwise
+     * @return raw pointer on the criterion interface, nullptr if the method fail
+     */
+    core::selection::criterion::CriterionInterface*
+    createInclusiveCriterion(const std::string& name,
+                             const core::selection::criterion::CriterionInterface::Values& values,
+                             std::string& error);
+
+    core::selection::criterion::CriterionInterface*
+    getSelectionCriterion(const std::string& strName);
```
dawagner: getCriterion ?

parameter/include/ParameterMgrPlatformConnector.h
```diff
@@ -54,12 +53,39 @@ class CParameterMgrPlatformConnector
...
+     */
+    core::selection::criterion::CriterionInterface*
+    createInclusiveCriterion(const std::string& name,
+                             const core::selection::criterion::CriterionInterface::Values& values,
+                             std::string& error);
+
     // Selection criterion retrieval
-    ISelectionCriterionInterface* getSelectionCriterion(const std::string& strName) const;
+    core::selection::criterion::CriterionInterface*
+    getSelectionCriterion(const std::string& strName) const;
```
dawagner: "getCriterion" ?

parameter/include/ParameterMgrFullConnector.h
```diff
@@ -71,10 +70,38 @@ class CParameterMgrFullConnector
...
+     *
+     * @param[in] name, the criterion name
+     * @param[in] values available value pairs for the created criterion
+     * @param[out] error the string describing the error if an error occurred
+     *                   undefined otherwise
+     * @return raw pointer on the criterion interface, nullptr if the method fail
+     */
+    core::selection::criterion::CriterionInterface*
+    createExclusiveCriterion(const std::string& name,
+                             const core::selection::criterion::CriterionInterface::Values& values,
```
dawagner: I'd prefer making it as easy+obvious for the user by simply taking a std::map<string, int> as argument.
You can make it even easier by overloading this method and take a std::list<string> (for people who don't care about the criterion numerical value).

parameter/SelectionCriterionRule.cpp
```diff
@@ -183,12 +170,14 @@ bool CSelectionCriterionRule::fromXml(const CXmlElement& xmlElement, CXmlSeriali
     // Get Value
     string strValue = xmlElement.getAttributeString("Value");
 
-    if (!_pSelectionCriterion->getCriterionType()->getNumericalValue(strValue, _iMatchValue)) {
+    int numericalValue;
+    if (!_pSelectionCriterion->getNumericalValue(strValue, numericalValue)) {
```
krocard: You are not handling the *none* case.

parameter/SelectionCriterionRule.cpp
```diff
@@ -96,21 +87,25 @@ bool CSelectionCriterionRule::parse(CRuleParser& ruleParser, string& strError)
...
+                   (_pSelectionCriterion->isInclusive() ? "inclusive" : "exclusive") +
+                   " criterion: " + _pSelectionCriterion->getCriterionName();
 
         return false;
     }
 
     // Value
-    if (!_pSelectionCriterion->getCriterionType()->getNumericalValue(strValue, _iMatchValue)) {
+    int numericalValue;
+    if (!_pSelectionCriterion->getNumericalValue(strValue, numericalValue)) {
```
krocard: Please handle the *none* case.

parameter/criterion/include/criterion/Criteria.h
```diff
@@ -0,0 +1,162 @@
...
+
+    /** Reset the modified status of criteria */
+    void resetModifiedStatus();
+
+    /** Xml Serialization method
+     *
+     * @param[out] xmlElement the current xml element to fill with data
+     * @param serializingContext context of the current serialization
+     */
+    virtual void toXml(CXmlElement& xmlElement, CXmlSerializingContext& serializingContext) const;
```
dawagner: override

parameter/criterion/include/criterion/Criterion.h
```diff
@@ -0,0 +1,204 @@
...
+     */
+    Criterion(const std::string& name, const Values& values, core::log::Logger& logger);
+
+    // @{
+    /** @see CriterionInterface */
+    virtual bool setState(const State& state, std::string& error) override;
+
+    virtual State getState() const override final;
+
+    virtual std::string getCriterionName() const override final;
```
dawagner: rename as "getCriterion" ?

clero: "getName" no ?

parameter/ConfigurableDomain.cpp
clero: done

parameter/criterion/include/criterion/Criterion.h
dawagner: indeed

parameter/criterion/include/criterion/client/CriterionInterface.h
```diff
@@ -30,17 +30,59 @@
...
+{
 
-class ISelectionCriterionInterface
+class CriterionInterface
 {
 public:
-    virtual void setCriterionState(int iState) = 0;
-    virtual int getCriterionState() const = 0;
+    /** Type which represent a criterion state part */
+    using Value = std::string;
```
krocard: Why is it defining in the client interface but never used (in the client interface) ?

bindings/c/ParameterFramework.cpp
clero: not needed any more

parameter/criterion/include/criterion/Criterion.h
```diff
@@ -0,0 +1,204 @@
...
+namespace selection
+{
+namespace criterion
+{
+
+/** Criterion object used to apply rules based on system state */
+class Criterion : public IXmlSource, public CriterionInterface
+{
+public:
+    using InvalidCriterionError = std::runtime_error;
```
krocard: doxygen please.

parameter/criterion/include/criterion/Criterion.h
```diff
@@ -0,0 +1,204 @@
...
+    bool hasBeenModified() const;
+
+    /** Reset the modified status of the criterion */
+    void resetModifiedStatus();
+
+    /** Request criterion state match with a desired method
+     *
+     * @param[in] method, the desired match method
+     * @param[in] state, the state to match
+     * @return true if the current state match the state given in parameter with the desired method
```
krocard: document that an exception is thrown if the method does not exist.

parameter/criterion/src/Criterion.cpp
```diff
@@ -0,0 +1,270 @@
...
+template<class MapType>
+MapType Criterion::gatherMap(MapType mapA, const MapType& mapB)
+{
+    mapA.insert(mapB.begin(), mapB.end());
+    return mapA;
+}
+
+template
+Criterion::MatchMethods Criterion::gatherMap<Criterion::MatchMethods>(MatchMethods mapA,
+                                                                      const MatchMethods& mapB);
```
krocard: line 2 long

parameter/criterion/src/Criterion.cpp
krocard: I do not like suggestion 2, it would duplicate the MatchMethods map in both class => The derived class would contain 2 classes for nothing.

parameter/criterion/src/Criterion.cpp
```diff
@@ -0,0 +1,270 @@
...
+    return mMatchMethods.at(method)(state);
+}
+
+bool Criterion::isMatchMethodAvailable(const std::string& method) const
+{
+    return mMatchMethods.count(method) == 1;
+}
+
+template<class MapType>
+MapType Criterion::gatherMap(MapType mapA, const MapType& mapB)
```
krocard: Could be a static method, no need to put it in the *.h*.

parameter/criterion/include/criterion/client/CriterionInterface.h
```diff
@@ -30,17 +30,59 @@
...
-#include "SelectionCriterionTypeInterface.h"
+namespace core
+{
+namespace selection
+{
+namespace criterion
+{
 
-class ISelectionCriterionInterface
+class CriterionInterface
```
krocard: doxygen ?

parameter/criterion/include/criterion/Criterion.h
```diff
@@ -0,0 +1,204 @@
...
+
+namespace core
+{
+namespace selection
+{
+namespace criterion
+{
+
+/** Criterion object used to apply rules based on system state */
+class Criterion : public IXmlSource, public CriterionInterface
```
krocard: non copyable ?

parameter/criterion/include/criterion/Criteria.h
```diff
@@ -0,0 +1,162 @@
...
+
+namespace core
+{
+namespace selection
+{
+namespace criterion
+{
+
+/** Criteria Handler */
+class Criteria : public IXmlSource
```
krocard: non copiable ?

parameter/criterion/src/Criteria.cpp
clero: done

parameter/include/ParameterMgrFullConnector.h
clero: done

parameter/include/ParameterMgrPlatformConnector.h
clero: done

parameter/criterion/include/criterion/Criterion.h
clero: done

parameter/criterion/include/criterion/Criterion.h
```diff
@@ -0,0 +1,204 @@
...
+     * Client need to be informed when the criterion has been modified
+     * and how many times.
+     * This method helps to count and log modifications.
+     */
+    void stateModificationsEvent();
+
+private:
+
+    /** Reset the criterion to its default state if needed */
+    inline bool setDefaultState();
```
krocard: inline ?

parameter/criterion/include/criterion/InclusiveCriterion.h
```diff
@@ -29,27 +29,39 @@
...
 
-    /**
-     * Add a new pair [integer, litteral] which represents a criterion
-     *
-     * @param[in] iValue integer value
-     * @param[in] strValue litteral value
-     * @param[out] strError string containing error information we can provide to client
-     * @return true if succeed false otherwise
+/** Criterion we can take several state values at the same time */
+class InclusiveCriterion : public Criterion
```
krocard: final

parameter/criterion/include/criterion/InclusiveCriterion.h
```diff
@@ -29,27 +29,39 @@
...
-    virtual ~ISelectionCriterionTypeInterface() {}
+    InclusiveCriterion(const std::string& name, const Values& values, core::log::Logger& logger);
+
+    // @{
+    /** @see CriterionInterface */
+    bool isInclusive() const override final;
+
+    bool setState(const State& state, std::string& error) override final;
+
+    std::string getFormattedState() const override final;
```
krocard: remove finals as is put on the class level.

bindings/c/ParameterFramework.h
```diff
@@ -156,25 +156,32 @@ const char *pfwGetLastError(const PfwHandler *handle) NONNULL;
 /** Set a criterion value given its name and value.
   * @param handle[in] @see PfwHandler
   * @param name[in] The name of the criterion that need to be changed.
-  * @param value If the criterion is exclusive, the index of the new value.
-  *              If the criterion is inclusive, a bit field where each bit
-  *              correspond to the value index.
+  * @param value If the criterion is exclusive, an array containing the new value.
+  *              If the criterion is inclusive, aan array containing all new values.
```
krocard: > aan

dawagner:
@clero I've rebased and merged part of this PR in #121 :v:

bindings/c/ParameterFramework.cpp
```diff
@@ -255,14 +249,14 @@ const char *pfwGetLastError(const PfwHandler *handle)
     return handle == NULL ? NULL : handle->lastStatus.msg().c_str();
 }
 
-static pfw::Criterion *getCriterion(const pfw::Criteria &criteria,
-                                    const string &name)
+static CriterionInterface *getCriterion(const pfw::Criteria &criteria,
+                                             const string &name)
```
krocard: Alignment

bindings/c/ParameterFramework.cpp
```diff
@@ -274,14 +268,14 @@ bool pfwSetCriterion(PfwHandler *handle, const char name[], int value)
         return status.failure("Can not set criterion \"" + string(name) +
                               "\" as the parameter framework is not started.");
     }
-    pfw::Criterion *criterion = getCriterion(handle->criteria, name);
+    CriterionInterface *criterion = getCriterion(handle->criteria, name);
```
krocard: CriterionInterface is so ugly... Can you not rename it Criterion ? :pray:

bindings/c/ParameterFramework.cpp
```diff
@@ -297,11 +291,26 @@ bool pfwGetCriterion(const PfwHandler *handle, const char name[], int *value)
...
-    *value = criterion->getCriterionState();
+    CriterionInterface::State state = criterion->getState();
+    *nbValue = state.size();
+
+    size_t i = 0;
+    for (auto &critValue : state) {
+        value[i] = strdup(critValue.c_str());
+        if (value[i] == nullptr) {
+            return status.failure("Can not get criterion " + string(name) +
+                                  " as there is no memory left");
```
krocard: Leak of previous strdup values !

bindings/c/ParameterFramework.cpp
```diff
@@ -274,14 +268,14 @@ bool pfwSetCriterion(PfwHandler *handle, const char name[], int value)
...
     if (criterion == NULL) {
         return status.failure("Can not set criterion " + string(name) + " as does not exist");
     }
-    criterion->setCriterionState(value);
-    return status.success();
+    CriterionInterface::State state{value, value + nbValue};
+    return status.forward(criterion->setState(state, status.msg()));
 }
-bool pfwGetCriterion(const PfwHandler *handle, const char name[], int *value)
+bool pfwGetCriterion(const PfwHandler *handle, const char name[], char *value[], size_t *nbValue)
```
krocard: How about returning a null terminated array ?

bindings/c/ParameterFramework.cpp
```diff
@@ -297,11 +291,26 @@ bool pfwGetCriterion(const PfwHandler *handle, const char name[], int *value)
...
     if (criterion == NULL) {
         return status.failure("Can not get criterion " + string(name) + " as it does not exist");
     }
-    *value = criterion->getCriterionState();
+    CriterionInterface::State state = criterion->getState();
+    *nbValue = state.size();
+
+    size_t i = 0;
+    for (auto &critValue : state) {
+        value[i] = strdup(critValue.c_str());
```
krocard: Where are you allocating the value array ?

bindings/c/ParameterFramework.h
```diff
@@ -156,25 +156,32 @@ const char *pfwGetLastError(const PfwHandler *handle) NONNULL;
...
-bool pfwSetCriterion(PfwHandler *handle, const char name[], int value) NONNULL USERESULT;
+bool pfwSetCriterion(PfwHandler *handle,
+                     const char name[],
+                     const char *value[],
+                     size_t nbValue) NONNULL USERESULT;
 /** Get a criterion value given its name.
-  * Same usage as pfwSetCriterion except that value is an out param.
+  * Same usage as pfwSetCriterion except that value and nbValue are out parameters.
   * Get criterion will return the last value setted with pfwSetCriterion independantly of pfwCommitCritenio.
+  * The callee MUST free the returned value array using pfwFree after use.
```
krocard: and also each string in the array (strdup'ed) !

bindings/c/Test.cpp
```diff
@@ -203,8 +203,8 @@ TEST_CASE_METHOD(Test, "Parameter-framework c api use") {
         }
         GIVEN("A criteria with lots of values")
         {
-            // Build a criterion with as many value as there is bits in int.
-            std::vector<char> names(sizeof(int) * CHAR_BIT + 1, 'a');
+            // Build a criterion with a lot of value
+            std::vector<char> names(300, 'a');
```
krocard: Why 300 ?  I think this test section is meaningless now (was testing the criterion value number limitation). Delete it ?

parameter/criterion/src/Criterion.cpp
clero: done

parameter/ParameterMgr.h
clero: This header still needed internally, maybe we can use a forward declaration as we are just returning a pointer on a Criterion object

parameter/criterion/src/Criterion.cpp
clero: removed

parameter/criterion/src/Criteria.cpp
clero: done

parameter/criterion/include/criterion/Criterion.h
clero: removed

parameter/criterion/include/criterion/Criteria.h
clero: already the case

parameter/criterion/include/criterion/InclusiveCriterion.h
clero: done

parameter/criterion/include/criterion/InclusiveCriterion.h
clero: done

parameter/criterion/include/criterion/client/CriterionInterface.h
clero: removed

parameter/criterion/src/Criterion.cpp
clero: done

parameter/criterion/include/criterion/Criteria.h
clero: done

parameter/criterion/src/Criterion.cpp
clero: done

parameter/SelectionCriterionRule.cpp
clero: done

parameter/SelectionCriterionRule.cpp
clero: done

parameter/criterion/include/criterion/Criterion.h
clero: done

parameter/criterion/include/criterion/Criterion.h
clero: done

parameter/criterion/src/Criterion.cpp
```diff
@@ -0,0 +1,255 @@
...
+}
+
+bool Criterion::match(const std::string& method, const State& state) const
+{
+    return mMatchMethods.at(method)(state);
+}
+
+bool Criterion::isMatchMethodAvailable(const std::string& method) const
+{
+    return mMatchMethods.count(method) == 1;
```
dawagner: suboptimal/strange semantic. proposal:

    return mMatchMethods.find(method) != mMatchMethods.end()

Same stuff for "isValueAvailable".

parameter/criterion/src/Criterion.cpp
clero: removed

parameter/criterion/src/Criterion.cpp
clero: moved to utility

parameter/criterion/src/Criterion.cpp
clero: why suboptimal? count exist in set and map only for this purpose no ?

bindings/c/ParameterFramework.h
clero: done

bindings/c/ParameterFramework.cpp
clero: done

bindings/c/ParameterFramework.cpp
clero: done

bindings/c/ParameterFramework.cpp
clero: in the nowhere land :8ball:  :)

bindings/c/ParameterFramework.h
clero: done

bindings/c/Test.cpp
clero: it shows all is working but yes we can remove it

bindings/c/ParameterFramework.cpp
clero: done

parameter/criterion/include/criterion/Criterion.h
clero: I don't now if non copyable is very interesting.
Concerning the renaming, I don't like it because Exclusive does not give many information. Moreover, the class is used internally as both inclusive and exclusive (due to the inheritance link).
I'm not fan of this link but it is necessary for now to avoid matching function copy paste. Moreover it is not very choking to say that a standard criterion can only have one value in its state and an inclusive one can have more.

parameter/criterion/src/Criterion.cpp
dawagner: ok, maybe it's not suboptimal for maps and sets... but I still find it strange to use "count" for checking an element's existence. I won't hold it against you if you decide to keep count() :)

bindings/c/ParameterFramework.cpp
clero: done

parameter/criterion/include/criterion/client/CriterionInterface.h
clero: doxygen done, you also want to remove it from criterion namespace ?

parameter/criterion/include/criterion/client/CriterionInterface.h
clero: It is needed for creation purpose (creation is made through ParameterMgr)
It cannot be defined in internal Criterion header as the client as no access to it, neither in the ParameterMgr. So I think this is the better place.

parameter/criterion/src/Criteria.cpp
clero: done

parameter/criterion/include/criterion/client/CriterionInterface.h
clero: done

parameter/SelectionCriterionRule.cpp
```diff
@@ -202,51 +193,11 @@ void CSelectionCriterionRule::toXml(CXmlElement& xmlElement, CXmlSerializingCont
...
-            _eMatchesWhen = (MatchesWhen)uiMatchesWhen;
-
-            return true;
-        }
-    }
-
-    strError = "Value not found";
-
-    return false;
+    xmlElement.setAttributeString("Value", *mMatchState.begin());
```
clero: need to handle "none" case

clero: done

parameter/criterion/include/criterion/client/Criterion.h
```diff
@@ -0,0 +1,83 @@
...
+    /** Criterion value collection
+     * When the criterion will be created, it will be checked that no internal values
+     * are duplicated (i.e that we have a set of value). Nevertheless, we decided to
+     * use a list to store values in order to let the user choose the value order.
+     * This is important as the first value will be chosen as default one.
+     * @see createExclusiveCriterion
+     */
+    using Values = std::list<Value>;
+    /** Criterion state representation */
+    using State = std::set<Value>;
```
dawagner: these `using`s could be moved outside of the class and directly in the core::criterion namespace; this would make it easier to use.

clero: done

PR #97, opened 2015-04-15
Turn off optimizations when building in Debug mode

clero:
CMake default debug flags contains contains only '-g' flag.
The optimization leads to coverage inconsistency when using
both clang and gcc compiler.
This patch adds '-O0' flag for Debug mode build to turn off
optimizations.


clero:
@dawagner @krocard please review

krocard:
:+1:

PR #98, opened 2015-04-15
travis: Use the docker infrastructure

dawagner:
As advertised by travis news:

    http://blog.travis-ci.com/2014-12-17-faster-builds-with-container-based-infrastructure/

using the new docker infra is faster and more reliable. This comes with the
cost of not being able to use "sudo", though. Installing packages is still
possible by specifying them in an "addons.apt" section in .travis.yml. We have
to remove "sudo make install" but we aren't using the installed libs and
binaries anyway.

I had some syntax issues with the "addons" and "env" sections and had to remove
the coverity subsection which wasn't configure for the 01org repository anyway.
I'm also removing the email notification since I never received any email and
actually don't want to.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -63,7 +63,5 @@ after_success:
           --gcov-options '\--long-file-names --preserve-paths'
 
 notifications:
-  email:
-    - david.wagner@intel.com
   irc:
     - "chat.freenode.net#parameter-framework"
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,41 +1,52 @@
+# Use travis docker infrastructure
+sudo: false
 language: cpp
-compiler:
-  - gcc
-#  - clang # not supported yet
 
-# Install a recent gcc, gcov and cmake,
+env:
+    global:
+        - PREFIX=$HOME/prefix
+
+compiler:
+    - gcc
+    # clang not supported yet
+    # - clang
+
+# Install a recent gcc and gcov,
 # it will not be necessary once travis worker is based on ubuntu > 12.04.
 # Install SWIG for bindings generation
-before_install:
-  - sudo add-apt-repository --yes ppa:ubuntu-toolchain-r/test
-  - sudo add-apt-repository --yes ppa:kalakris/cmake # Non official cmake backport
-  - sudo apt-get update -qq
-  - sudo pip install cpp-coveralls
+# Install valgrind for memcheck tests
+addons:
+    apt:
+        sources:
+            - ubuntu-toolchain-r-test
+        packages:
+            - swig
+            - valgrind
+            - g++-4.8
 
 install:
-  - sudo apt-get install --yes swig cmake valgrind g++-4.8
-  - sudo update-alternatives --install /usr/bin/g++ g++ /usr/bin/g++-4.8 100
-  - sudo update-alternatives --install /usr/bin/gcov gcov /usr/bin/gcov-4.8 100
-  - sudo wget --directory-prefix /usr/include/
+    - pip install --user cpp-coveralls; export PATH=$HOME/.local/bin:$PATH
+    - wget --directory-prefix $PREFIX/include
               https://raw.github.com/philsquared/Catch/master/single_include/catch.hpp
+
+before_script:
+    - if [ "$CC" = "gcc" ]; then export CC=gcc-4.8 CXX=g++-4.8; fi
 
 # how to build
 script:
   - ( mkdir build_debug && cd build_debug &&
-        cmake -DCMAKE_BUILD_TYPE=Debug -DCOVERAGE=ON .. &&
+        cmake -DCMAKE_PREFIX_PATH=$PREFIX -DCMAKE_BUILD_TYPE=Debug -DCOVERAGE=ON .. &&
         make -j &&
         CTEST_OUTPUT_ON_FAILURE=1 make ExperimentalTest ExperimentalMemCheck )
   - ( mkdir build && cd build &&
-        cmake .. &&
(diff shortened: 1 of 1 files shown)


dawagner:
@krocard please review

gdenneul:
:+1:

.travis.yml
```diff
@@ -1,41 +1,55 @@
...
+    global:
+        - PREFIX=$HOME/prefix
+
 compiler:
-  - gcc
-#  - clang # not supported yet
+    - gcc
+    - clang
+
+matrix:
```
krocard: useless matrix, why did you not kept clang commented ?

.travis.yml
```diff
@@ -1,41 +1,55 @@
...
+    allow_failures:
+        # clang not supported yet
+        - compiler: clang
```
krocard: useless matrix, why did you not kept clang commented ?

.travis.yml
```diff
@@ -1,41 +1,55 @@
...
 
 # Install a recent gcc, gcov and cmake,
```
krocard: cmake are not longer updated.

.travis.yml
```diff
@@ -1,41 +1,55 @@
...
-  - sudo apt-get install --yes swig cmake valgrind g++-4.8
-  - sudo update-alternatives --install /usr/bin/g++ g++ /usr/bin/g++-4.8 100
-  - sudo update-alternatives --install /usr/bin/gcov gcov /usr/bin/gcov-4.8 100
-  - sudo wget --directory-prefix /usr/include/
+    - pip install --user cpp-coveralls; export PATH=$HOME/.local/bin:$PATH
+    - wget --directory-prefix $PREFIX
               https://raw.github.com/philsquared/Catch/master/single_include/catch.hpp
 
+before_script:
+    - if [ "$CC" = "gcc" ]; then export CC=gcc-4.8 CXX=g++-4.8; fi
```
krocard: export CC=${CC/$gcc$/gcc-4.8}
export CXX=${CC/$g++$/g++-4.8}

.travis.yml
```diff
@@ -1,41 +1,55 @@
...
+            - valgrind
+            - g++-4.8
 
 install:
-  - sudo apt-get install --yes swig cmake valgrind g++-4.8
-  - sudo update-alternatives --install /usr/bin/g++ g++ /usr/bin/g++-4.8 100
-  - sudo update-alternatives --install /usr/bin/gcov gcov /usr/bin/gcov-4.8 100
-  - sudo wget --directory-prefix /usr/include/
+    - pip install --user cpp-coveralls; export PATH=$HOME/.local/bin:$PATH
+    - wget --directory-prefix $PREFIX
```
krocard: > $ man cmake
> ***CMAKE_PREFIX_PATH***
>   Path used for searching by FIND_XXX(), with appropriate suffixes added.
> 
>   Specifies a path which will be used by the  FIND_XXX()  commands.  It  contains  the  "base"  directories,  the  FIND_XXX()  commands  append appropriate subdirectories to the base directories. So FIND_PROGRAM() adds /bin to each of the directories in the path, FIND_LIBRARY() appends /lib to each of the  directories,  and  FIND_PATH()  and  **FIND_FILE()  append  /include**  . By default it is empty, it is intended to be set by the project.

Catch is found with FIND_FILE. If I understand correctly the manual `cmake` looks for `catch.hpp` in `${CMAKE_PREFIX_PATH}/include`. Should you not download the header to `$PREFIX/include` ?

.travis.yml
```diff
@@ -48,26 +62,9 @@ after_success:
           --exclude "test/test-subsystem"
           --exclude "bindings/c/Test.cpp"
           --exclude "test/tokenizer"
+          --gcov /usr/bin/gcov-4.8
           --gcov-options '\--long-file-names --preserve-paths'
 
 notifications:
-  email:
-    - david.wagner@intel.com
```
krocard: How is this related to docker ?

.travis.yml
```diff
@@ -48,26 +62,9 @@ after_success:
...
   irc:
     - "chat.freenode.net#parameter-framework"
-
```
krocard: How is this change related to docker ?

.travis.yml
dawagner: oops, you are right; that's what I intended to do.

.travis.yml
dawagner: see commit message

.travis.yml
dawagner: see commit message... but you are right, in theory I should have done it in a separate patch.....

.travis.yml
dawagner: done

.travis.yml
dawagner: done

.travis.yml
dawagner: done

krocard:
:+1:

.travis.yml
```diff
@@ -1,41 +1,52 @@
...
+        cmake -DCMAKE_PREFIX_PATH=$PREFIX -DCMAKE_INSTALL_PREFIX=../install .. &&
         make -j &&
-        sudo make install &&
-        sudo ldconfig &&
-        CTEST_OUTPUT_ON_FAILURE=1 make test )
+        CTEST_OUTPUT_ON_FAILURE=1 make test &&
+        make install)
   - ( cd skeleton-subsystem &&
-        cmake . &&
+        cmake -DCMAKE_INSTALL_PREFIX=../install . &&
```
OznOg: I usually prefer global path for this

PR #99, opened 2015-04-15
xmlcoverage: be able to ignore changes to unknown criteria

dawagner:
Since a parameter-framework client may set a criterion before calling start(),
the logs may show changes to criteria before the coverage script knows the list
of criteria. It used to cause an unrecoverable error even with the "--force"
option.

This option now makes the coverage script ignore such errors. The script
already takes the initial values of criteria into account when the
parameter-framework starts, so there isn't any information loss.

However, if this error happens for any other error than described in the first
paragraph, you probably don't want to silence it.

### DIFF
--- a/tools/coverage/coverage.py
+++ b/tools/coverage/coverage.py
@@ -720,6 +720,14 @@ class ParsePFWlog():
     MATCH = "match"
     ACTION = "action"
 
+    class ChangeRequestOnUnknownCriterion(CustomError):
+        def __init__(self, criterion):
+            self.criterion = criterion
+
+        def __str__(self):
+            return ("Change request on an unknown criterion %s." %
+                self.criterion)
+
     def __init__(self, domains, criteria, ErrorsToIgnore=()):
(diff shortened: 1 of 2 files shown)


dawagner:
@krocard @gdenneul please review

krocard:
:+1:

tools/coverage/coverage.py
```diff
@@ -1003,6 +1021,9 @@ def __init__(self):
                 errorToIgnore.append(Configuration.IneligibleConfigurationAppliedError)
 
             if options.incoherentCriterionFlag:
+                errorToIgnore.append(ParsePFWlog.ChangeRequestOnUnknownCriterion)
```
gdenneul: wtf?

gdenneul:
:-1:

tools/coverage/coverage.py
dawagner: This will be fixed by #101

PR #101, opened 2015-04-20
xml coverage: fix a copy-paste mistake

dawagner:
Commit ccb164c15d0f46fb233b67075f37bede8f8ad855 introduced the possibility to
ignore changes to unknown criterion through an option passed to coverage.py but
a copy-paste mistake swapped the action of this option with the option ignoring
incoherent criterion state changes.

Both flags are set by 'aplog2coverage.sh -f' so the mistake went unnoticed but
it must be fixed nevertheless.


OznOg:
:+1:

PR #102, opened 2015-04-20
Allows 64 values for inclusive criterion

clero:
Current implementation allows only 32 values for an inclusive
criterion. It comes from the use of a 32 bit integer for the
internal state representation.
This patch replaces the state representation by a 64 bit integer
which allows 64 bit. This is a temporary patch as the entire
interface should be reworked to avoid such problems in the
future.

Change-Id: I0521569a0a9cd40d83b75fe5c046bbf48c155a07
Tracked-On: https://jira01.devtools.intel.com/browse/IMINAN-31110

### DIFF
--- a/bindings/python/pfw.i
+++ b/bindings/python/pfw.i
@@ -198,11 +198,11 @@ class ISelectionCriterionTypeInterface
 %}
 
 public:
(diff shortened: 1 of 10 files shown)


krocard:
Pull request is deprecated due to the refactor in #96.

PR #104, opened 2015-04-22
android/python bindings: force gcc and 32bits compilation

dawagner:
Since Android only provides a 32bits version of the python interpreter and
libs, we need to compile for 32bits even on 64bits hosts.

Also, since it seems that clang is the default compiler for host, we need to
explicitely forbid it as long as the parameter-framework is compiled with gcc.

Change-Id: I5d484c0e58c304680192f443acd81258de179e52

### DIFF
--- a/bindings/python/Android.mk
+++ b/bindings/python/Android.mk
@@ -38,6 +38,11 @@ include $(CLEAR_VARS)
 LOCAL_MODULE := _PyPfw
 
 LOCAL_CPP_EXTENSION := .cxx
+# As long as the parameter-framework is compiled with gcc, we must avoid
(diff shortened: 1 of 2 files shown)


dawagner:
@OznOg @krocard please review

bindings/python/Android.mk
```diff
@@ -38,6 +38,8 @@ include $(CLEAR_VARS)
 LOCAL_MODULE := _PyPfw
 
 LOCAL_CPP_EXTENSION := .cxx
+LOCAL_CLANG := false
```
OznOg: explain why false

OznOg:
:+1:

PR #105, opened 2015-04-23
python bindings: activate thread-safety code generation

dawagner:
The Parameter Framework may spawn a thread listening to request for command
execution. In this case, whenever a command produces a log, a callback function
provided by the client is called. When the client is a python script using the
python bindings, this causes a python function to be called in a different
context than the main thread. This can only work safely if a specific lock is
held before hand; see

    https://docs.python.org/2/c-api/init.html#non-python-created-threads

Fortunately, SWIG can generate the corresponding code automatically using

    %module(threads=1, ...) mymodule

### DIFF
--- a/bindings/python/pfw.i
+++ b/bindings/python/pfw.i
@@ -32,7 +32,7 @@
 // like derived C++ classes (calls to virtual methods will be properly
 // forwarded to Python) - only on classes for which is it specified, see
 // ILogger below..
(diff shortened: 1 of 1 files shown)


dawagner:
Thanks @krocard for the link mentioned in the commit message.

krocard:
:+1:

PR #109, opened 2015-05-04
Check criterion numerical value validity if it is inclusive

zchokri:
In case of Inclusive Criterion, its appropriate numerical value may
be equal to the addition of several values which results to non valid
numerical values.

This patch checks the criterion value validity of inclusive criterions.

Signed-off-by: Chokri ZAGROUBA <chokrix.zagrouba@intel.com>


krocard:
This pull request has been pending for a long time. What is the requirement fullfiled by it ? Why is a new public api added ? Is their still a need for it ?

krocard:
Will close in a couple of weeks if their is no answer.

PR #111, opened 2015-05-07
Stop using stlport in Android.

DanAlbert:
Stlport has been deprecated in Android for a while now, and is finally
being removed very soon.

### DIFF
--- a/parameter/Android.mk
+++ b/parameter/Android.mk
@@ -125,7 +125,8 @@ common_cflags := \
         -Wall \
         -Werror \
         -Wextra \
-        -Wno-unused-parameter
+        -Wno-unused-parameter \
+        -Wno-maybe-uninitialized \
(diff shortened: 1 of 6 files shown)


parameter/Android.mk
```diff
@@ -125,7 +125,8 @@ common_cflags := \
         -Wall \
         -Werror \
         -Wextra \
-        -Wno-unused-parameter
+        -Wno-unused-parameter \
+        -Wno-maybe-uninitialized \
```
krocard: Why is this needed ? Is it linked to the new lib std  ?

krocard:
We will merge that after the next android release as this patch on android-L breaks the compilation (no default stl).

dawagner:
Done in tag v2.6.0+no-stlport.

PR #29, opened 2014-12-05
Check libc call failure

krocard:
The parameter framework was not compiling on chrome because the libc is forcing call status to be checked  

Handle failure of calls by returning an error if possible or just logging if not.

PR #113, opened 2015-05-14
Fix unused result warnings.

DanAlbert:
Android has enabled _FORTIFY_SOURCE for our host builds now, which
enables the unused result attributes on a number of functions in the C
library. This change quiets those warnings.

The result from pipe(2) is now properly plumbed to report errors, but
read(2) and write(2) only handle EINTR now; any real errors will still
be a problem. I haven't done any more because I was unsure about how
errors should be handled in those cases.


remote-processor/RemoteProcessorServer.cpp
```diff
@@ -44,8 +44,12 @@ using std::string;
 CRemoteProcessorServer::CRemoteProcessorServer(uint16_t uiPort, IRemoteCommandHandler* pCommandHandler) :
     _uiPort(uiPort), _pCommandHandler(pCommandHandler), _bIsStarted(false), _pListeningSocket(NULL), _ulThreadId(0)
 {
+}
+
+bool CRemoteProcessorServer::init()
+{
     // Create inband pipe
-    pipe(_aiInbandPipe);
+    return pipe(_aiInbandPipe) == 0;
```
dawagner: you could do that in the "start()" method instead of adding a new method. See https://github.com/01org/parameter-framework/pull/29/files were work has started for checking the return values of libc calls.

dawagner:
Thanks for doing this :)

Please rebase and resubmit against the `android-m` branch on my fork: https://github.com/dawagner/parameter-framework/compare/android-m...DanAlbert:fix-unused-results : I've pushed a change in the CMakefile in order to switch to c++11, which should make Travis happy w.r.t. `auto` and `nullptr`.

krocard:
@dawagner Much of this is covered by #29.

PR #114, opened 2015-05-15
documentation: add a Doxygen file and target

dawagner:
These can be used to regenerate the documentation found at

    http://01org.github.io/parameter-framework/

There is a new "doc" make target (not built by default).


CMakeLists.txt
```diff
@@ -36,6 +36,11 @@ endif(COMMAND cmake_policy)
 
 project(parameter-framework)
 
+# find and set the Parameter Framework's version
+execute_process(COMMAND git describe --tags --abbrev=0
```
krocard: I would rather have the exact version, because (one patch over v1) != (v1). => remove --abbrev=0
It would also be good to know if the tree is dirty. => add --dirty

~~~
  git describe --tags --dirty
~~~

doc/CMakeLists.txt
```diff
@@ -0,0 +1,48 @@
...
+# ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+# (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+# LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON
+# ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+# (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+# SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+find_package(Doxygen)
+
+if(NOT DOXYGEN_FOUND)
```
krocard: I would rather have an option, default to on than disable the feature if the dependency is not met. => explicit build requirements.
 - add REQUIRE to the find package
 - do not test DOXYGEN(_DOT)?_FOUND

krocard:
:-1:

CMakeLists.txt
dawagner: done

doc/CMakeLists.txt
dawagner: done in a slightly different way.

krocard:
:+1:

PR #117, opened 2015-05-20
Xml Serialization: Manage the Unit and Description fields

cc6565:
This patchs adds the support of Unit and Description fields for the import / export and showProperties commands.

### DIFF
--- a/parameter/Element.cpp
+++ b/parameter/Element.cpp
@@ -39,6 +39,8 @@
 
 using std::string;
 
+const std::string CElement::gDescriptionPropertyName = "Description";
+
 CElement::CElement(const string& strName) : _strName(strName), _pParent(NULL)
 {
 }
@@ -200,6 +202,14 @@ void CElement::showProperties(string& strResult) const
 {
     strResult = "\n";
     strResult += "Kind: " + getKind() + "\n";
+    showDescriptionProperty(strResult);
+}
+
+void CElement::showDescriptionProperty(std::string &strResult) const
+{
+    if (!getDescription().empty()) {
+        strResult += gDescriptionPropertyName + ": " + getDescription() + "\n";
+    }
 }
 
 // Conversion utilities
@@ -249,6 +259,8 @@ void CElement::logValue(string& strValue, CErrorContext& errorContext) const
 // From IXmlSink
 bool CElement::fromXml(const CXmlElement& xmlElement, CXmlSerializingContext& serializingContext)
 {
+    setDescription(getXmlDescriptionAttribute(xmlElement));
+
     // Propagate through children
     CXmlElement::CChildIterator childIterator(xmlElement);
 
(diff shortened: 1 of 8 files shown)


parameter/Element.cpp
```diff
@@ -39,6 +39,8 @@
 
 using std::string;
 
+const std::string& CElement::gDescriptionPropertyName = "Description";
```
krocard: You are using a rather obscure feature of c++: binding a const reference to a rvalue extends it's lifetime to that of the reference.
Are you sure you want to do that ? I would rather remove the `&`.

parameter/ParameterType.cpp
```diff
@@ -36,6 +36,8 @@
 
 using std::string;
 
+const std::string& CParameterType::gUnitPropertyName = "Unit";
```
krocard: idem, remove the `&`.

parameter/Element.cpp
```diff
@@ -249,6 +259,8 @@ void CElement::logValue(string& strValue, CErrorContext& errorContext) const
 // From IXmlSink
 bool CElement::fromXml(const CXmlElement& xmlElement, CXmlSerializingContext& serializingContext)
 {
+    setDescription(xmlElement.getAttributeString(gDescriptionPropertyName));
```
krocard: setDescription(getXmlDescriptionAttribute(xmlElement)

parameter/Subsystem.cpp
```diff
@@ -104,6 +104,8 @@ bool CSubsystem::needResync(bool bClear)
 // From IXmlSink
 bool CSubsystem::fromXml(const CXmlElement& xmlElement, CXmlSerializingContext& serializingContext)
 {
+    setDescription(getXmlDescriptionAttribute(xmlElement));
```
krocard: why is it done here ? it is already done in element.

parameter/TypeElement.cpp
```diff
@@ -76,8 +76,9 @@ bool CTypeElement::hasMappingData() const
 // Element properties
 void CTypeElement::showProperties(std::string& strResult) const
 {
-    (void)strResult;
-    // Prevent base from being called in that context!
+    showDescriptionProperty(strResult);
+    // Prevent base from being called from the Parameter Type context.
```
krocard: what does the parameter type has to do here ? TypeElement does not know Parameter Type. Why do we need to override showProperties ? Element already print the description.

krocard:
:-1:

parameter/TypeElement.cpp
krocard: The name attribute is duplicated in the type and instance (common base Element). As a result we do not want to forwarded to the base class because it would duplicate the name attribute.

parameter/Subsystem.cpp
krocard: The Element::fromXml is not called because we subsystem does not use the generic fromXml algorithm.

parameter/TypeElement.cpp
cc6565: Even if the description belongs to element, the field is appended in most cases to the parametertype and not to the parameter class itself. So, we had to forward only the retrieval of the description property field to minimize the changes. To avoid this, we shall have attached the description field to the parameter class properly, which might be more tricky.

parameter/Subsystem.cpp
cc6565: Nope, for the subsystem, the fromXml implementation of Element is not called.

parameter/Element.h
```diff
@@ -144,6 +144,30 @@ class CElement : public IXmlSink, public IXmlSource
 
     // Class kind
     virtual std::string getKind() const = 0;
+
+    /**
+     * Fill the Description field of the Xml Element during XML composing.
+     *
+     * @param[in|out] xmlElement to fill with the description
```
dawagner: in,out

cc6565:
@louizo test

krocard:
:+1:

PR #119, opened 2015-05-28
libremote-processor must link against libpfw_utility

dawagner:
It uses Utility::fullRead and Utility::fullWrite and must therefore link
against libpfw_utility.


dawagner:
@krocard please review.

krocard:
:+1:

PR #120, opened 2015-05-28
ParameterMgr: make the logger mutable

dawagner:
If it isn't mutable, we can't possibly log from const methods.


krocard:
:+1:

clero:
:+1:

PR #126, opened 2015-06-01
Criterion rework step 2

dawagner:
Second step of the "criterion rework".

Warning: This step breaks the client's criterion API.

Instead of creating a criterion type (which defined the kind, inclusive or exclusive), then adding ValuePairs and then "instantiating" a criterion with that type, the client now simply create a criterion in 2 steps: each kind has its own method (createExclusiveCriterion/createInclusiveCriterion) and takes a name as argument. The client than adds ValuePairs to this criterion before starting the parameter framework.

More simplification to come.


dawagner:
@clero please review.

dawagner:
license updates done.

clero:
:+1:

PR #130, opened 2015-06-03
Fix ILogger::operator<< when called on list<string>

### LINKED ISSUE #125: LogWrapper::<<(list) does not behave like LogWrapper::<<(string) used on each element of the list
Calling `info() << someList` should be identical to calling `info() << element` on each element of `someList`.

This is not the case because when used on a list, the string is concatenated with `\n` and the logging callback is called once. Instead, the logging  callback should be called for each line and we shouldn't assume anything on the "newline" character.

dawagner:
Calling info() << someList should be identical to calling info() << element on
each element of someList.

This was not the case because when used on a list, the string was concatenated
with \n and the logging callback called once. Instead, the logging callback
is now called for each line.

Fixes #125.

### DIFF
--- a/parameter/log/include/log/LogWrapper.h
+++ b/parameter/log/include/log/LogWrapper.h
@@ -71,17 +71,17 @@ public:
     /** Class destructor */
     ~LogWrapper()
     {
-        if (!mLog.str().empty()) {
-            if(isWarning) {
-                mLogger.warning(mProlog + mLog.str());
-            } else {
-                mLogger.info(mProlog + mLog.str());
-            }
+        for (auto& log : mLogs) {
+            (mLogger.*mLogCallback)(mProlog + log.str());
         }
     }
 
     /**
      * Simulate stream behaviour
+     *
+     * Both operator<<() versions must not be used in conjunction: either use
+     * the 'scalar' version (this one) or the 'list' version but not both on
+     * the same object.
      *
      * @tparam T the type of the information to log
      * @param[in] log the information to log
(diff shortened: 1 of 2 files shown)


dawagner:
@krocard @clero @OznOg please review.

dawagner:
Build breaks because the skeleton plugin is not explicitly compiled with C++11 and ILogger.h now requires it.

parameter/log/include/log/LogWrapper.h
```diff
@@ -71,11 +71,11 @@ class LogWrapper
...
     {
-        if (!mLog.str().empty()) {
-            if(isWarning) {
-                mLogger.warning(mProlog + mLog.str());
-            } else {
-                mLogger.info(mProlog + mLog.str());
+        auto logfunc = isWarning ? &ILogger::warning : &ILogger::info;
+
+        for (auto& log : mLogs) {
+            if (!log.str().empty()) {
```
krocard: Are you sure this empty test is still useful ? A stream string is only created if there is things to put in.

dawagner: right.

parameter/log/include/log/LogWrapper.h
```diff
@@ -83,33 +83,38 @@ class LogWrapper
...
      */
     template <class T>
     LogWrapper& operator<<(const T &log)
     {
-        mLog << log;
+        if (mLogs.empty()) {
+            mLogs.emplace_back();
+        }
+
+        mLogs.back() << log;
```
krocard: maybe we could add std::to_string instead of an ostringstream ?
... Never mind, `std::to_string(std::string)` <=> `id` does not exist. Unfortunately.

parameter/log/include/log/LogWrapper.h
```diff
@@ -71,45 +71,48 @@ class LogWrapper
     /** Class destructor */
     ~LogWrapper()
     {
-        if (!mLog.str().empty()) {
-            if(isWarning) {
-                mLogger.warning(mProlog + mLog.str());
-            } else {
-                mLogger.info(mProlog + mLog.str());
-            }
+        auto logfunc = isWarning ? &ILogger::warning : &ILogger::info;
```
clero: as it depends on the template argument, shouldn't logfunc be static?

clero:
:+1:

PR #133, opened 2015-06-05
Build on darwin in the android framework with clang

krocard:
In order to build the parameter-framework on darwin (osx), clang must be supported.
 - Fix the code to compile with clang
 - Enable travis's clang build
 - No longer forbid to build with clang in android.mk

Notes

 - the parameter framework still does not support out of the box compilation on darwin due to warnings (deprecated features mostly) that are considered as errors.
 - the first patch removed the fake parameter-framework_include lib that was used by the plugins to find PFw headers. This mean that all plugin makefiles must be updated.


remote-processor/Android.mk
```diff
@@ -79,10 +75,8 @@ include $(CLEAR_VARS)
 
 LOCAL_SRC_FILES := $(common_src_files)
 
-LOCAL_STATIC_LIBRARIES := libpfw_utility_host
```
dawagner: I think this line shouldn't be removed.

remote-processor/Socket.cpp
```diff
@@ -50,6 +51,19 @@ CSocket::CSocket() : _iSockFd(socket(AF_INET, SOCK_STREAM, 0))
...
+#   if not defined(SIGPIPE)
+        // Pipe signal does not exist, there no sigpipe to ignore on send
+#   elif defined(SO_NOSIGPIPE)
+        const int set = 1;
+        setsockopt(sd, SOL_SOCKET, SO_NOSIGPIPE, &set, sizeof(set));
+#   elif defined(MSG_NOSIGNAL)
+        // Use flag NOSIGNAL on send call
+        mSendFlag = MSG_NOSIGNAL;
+#   else
+#       error Can no disable SIGPIPE
```
dawagner: Y U make a typo ? :p

support/android/build_pfw_settings.mk
```diff
@@ -35,7 +35,8 @@ $(LOCAL_BUILT_MODULE): MY_TOOL := $(HOST_OUT)/bin/domainGenerator.py
 # As of Android K, python is available as prebuilt. We can't reliably use the
 # host's default python because the low-level python binding has been compiled
 # against Android's Python headers.
-$(LOCAL_BUILT_MODULE): MY_PYTHON := prebuilts/python/linux-x86/2.7.5/bin/python
+# BTW, python is only available in 32bits for now, thus arch is forces to 32bits
```
dawagner: forced

remote-processor/Socket.cpp
krocard: s/no/not

krocard:
:+1: @dawagner @OznOg Please review. These are essentially your patches a bit reworked (added comments) for some.

bindings/python/Android.mk
```diff
@@ -74,6 +77,26 @@ LOCAL_GENERATED_SOURCES := $(generated-sources-dir)/pfw_wrap.cxx $(generated-sou
...
+    # Contrary to linux, on darwin, a python 64 bit executable is installed
+    # in the x86 prebuild directory,
+    # As all host libraries are 32 bit in android. We can not link and host
+    # python module against the prebuild python library.
+    #
+    # As a *dirty* workaround, use the system's python configuration and hope
+    # it will be compatible with the prebuild python interpreter used at runtime.
+    # To summarize the prebuild python (64 bit?) interpreter will load a
+    # python native module (32bit) linked with the host (32 bit ?) python library.
+    LOCAL_LDLIBS += $(shell python-config --ldflags)
```
dawagner: It looks like this variable should be LOCAL_LDFLAGS.

krocard: The only differences between LDFLAGS and LDLIBS is there place in ld command line. As you can see in build/core/definitions.mk L1370 and L1380 LDFLAGS is added athe be begining and LDLIBS is added at the end.
As python-config flags must have precedence, add them at the end.

PR #136, opened 2015-06-09
Android: Make libxmserializer dynamic

krocard:
The Parameter Framework has recently removed (this trick was not working on osx) its *_include libraries that were used to include headers without linking against the libraries in the android build system.

There are two ways to include library headers:
 - copy the headers (but that would be in contradiction with how the main pfw headers are included)
 - link against the library (chosen solution)

In order for the plugins to avoid linking against a static
libxmlserializer - ie add it and its dependency to all plugins -
the xmlserializer library is now dynamic.

The [alsa](https://github.com/01org/parameter-framework-plugins-alsa/pull/15) and [filesystem](https://github.com/01org/parameter-framework-plugins-filesystem/pull/8) plugins are also updated.

### DIFF
--- a/bindings/python/Android.mk
+++ b/bindings/python/Android.mk
@@ -44,8 +44,7 @@ LOCAL_CLANG := false
 # Android only provides a 32bit version of python.
 LOCAL_32_BIT_ONLY := true
 
-LOCAL_SHARED_LIBRARIES := libparameter_host
-LOCAL_STATIC_LIBRARIES := libxmlserializer_host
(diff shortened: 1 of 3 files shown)


krocard:
@OznOg @dawagner Please review this pull request and the 2 attached.

OznOg:
:+1:

dawagner:
Same commit message comment.

Please check that it it compiling on a android nexus lunch target (or - easier - for the arm emulator): I've noted differences between arm and x86 linking phase in the past.

PR #137, opened 2015-06-15
doc: some directories were missing from doxygen's list

dawagner:
A line ending in the INPUT variable (in Doxyfile.in) was not correctly escaped,
which resulted in several directories to be ignored. Let's add them and also
activate the RECURSIVE option in order to parse subdirectories.


dawagner:
@krocard please review

krocard:
:+1:

PR #138, opened 2015-06-17
Remove mappingdata xmlsink inheritance

dawagner:
The MappingData inheritance from XmlSink can be avoided at no cost.


dawagner:
@clero please review

clero:
:+1:

PR #140, opened 2015-06-18
travis: exclude 'install' directory from coverage PATH

clero:
The directory used to install the PF was covered. Thus,
include files were covered twice.
This patch removes this directory from the coverage PATH.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -54,6 +54,7 @@ after_success:
           --exclude "build_debug/bindings/python"
           --exclude "build_debug/CMakeFiles"
           --exclude "build"
+          --exclude "install"
           --exclude "skeleton-subsystem"
           --exclude "test/test-subsystem"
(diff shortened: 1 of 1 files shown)


clero:
It seems to have no effect as installed header were not relevant for coverage purpose.
It still better as files appear only once.

krocard:
:+1:

krocard:
@dawagner Please merge if OK

dawagner:
I'd rather have the build and install directories outside of the tree: if we add another directory in the build, we'll have to add another exclusion. Instead, if we prevent making the source directory dirty, we can remove most exclusion rules.

krocard:
Fully agree to build out of source repository. Need to check if coveralls can handle it.

dawagner:
My approach didn't work. I'm cherry-picking your patch on #147.

PR #143, opened 2015-06-30
Fix default builder leak and make a warning non fatal in criterion's test.

krocard:
Catch parses expression to detect operators at build time by overloading operators. This leaded gcc to warn that some expression might be ambiguous after macro extension.

Port C++14's make_unique.

Manage default builders with unique_ptr to guard again leaks at compile time.


clero:
5c1a5cd
:-1:

clero:
:+1:

PR #146, opened 2015-07-09
Add check for inclusive criterion overflow

dawagner:
Inclusive criteria have a limited value number due to their internal state
representation. Currently, the only check made for this is at test-platform
level.

This patch adds a new check to make addValuePair API fail if too many values
are added to an inclusive criterion type.  A log has also been added to warn
the client.  The check at test-platform level has been removed.

Change-Id: Ie9c9ec8fb8561f949bb62adbab127bc900aa254b

### DIFF
--- a/parameter/SelectionCriterionType.cpp
+++ b/parameter/SelectionCriterionType.cpp
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2011-2014, Intel Corporation
+ * Copyright (c) 2011-2015, Intel Corporation
  * All rights reserved.
  *
  * Redistribution and use in source and binary forms, with or without modification,
@@ -30,6 +30,8 @@
 #include "SelectionCriterionType.h"
 #include "Tokenizer.h"
 
+#include <climits>
+
 #define base CElement
 
 const std::string CSelectionCriterionType::_strDelimiter = "|";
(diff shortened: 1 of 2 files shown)


dawagner:
@krocard @cc6565 please review.

parameter/SelectionCriterionType.cpp
```diff
@@ -51,6 +53,18 @@ std::string CSelectionCriterionType::getKind() const
 // From ISelectionCriterionTypeInterface
 bool CSelectionCriterionType::addValuePair(int iValue, const std::string& strValue)
 {
+    // An inclusive criterion is implemented as a bitfield over an int and
+    // thus, can't have values larger than the number of bits in an int.
```
krocard: This comment is not correct. How about:
~~~ cpp
 // Inclusive criterion is implemented as a bitfield over an int.
 // As such fitfield are usually constructed by shifting a bit, forbid the sign bit to be used.
 // Shifting a bit over the sign bit returns in an overflow, which is an undefined behavior.
~~~

parameter/SelectionCriterionType.cpp
```diff
@@ -51,6 +53,18 @@ std::string CSelectionCriterionType::getKind() const
...
+    static const unsigned int inclusiveCriterionMaxValue = 1 << (sizeof(iValue) * CHAR_BIT - 1);
+
+    if (_bInclusive && (unsigned int)iValue > inclusiveCriterionMaxValue) {
```
krocard: Would it not be equivalent to check `< 0` ?

test/test-platform/TestPlatform.cpp
```diff
@@ -333,14 +333,6 @@ bool CTestPlatform::createInclusiveSelectionCriterionFromStateList(
...
 
     uint32_t uiNbStates = remoteCommand.getArgumentCount() - 1;
-
-    if (uiNbStates > 32) {
-
-        strResult = "Maximum number of states for inclusive criterion is 32";
-
-        return false;
-    }
-
```
krocard: This mean that `0x1 << uiState` might overflow => UB ! Maybe we could replace 0x1 => 1U.

test/test-platform/TestPlatform.cpp
```diff
@@ -397,13 +389,6 @@ bool CTestPlatform::createInclusiveSelectionCriterion(const string& strName,
     ISelectionCriterionTypeInterface* pCriterionType =
         _pParameterMgrPlatformConnector->createSelectionCriterionType(true);
 
-    if (uiNbStates > 32) {
-
-        strResult = "Maximum number of states for inclusive criterion is 32";
-
-        return false;
-    }
-
```
krocard: Idem add U

parameter/SelectionCriterionType.cpp
dawagner: obsolete

parameter/SelectionCriterionType.cpp
dawagner: obsolete

test/test-platform/TestPlatform.cpp
dawagner: done in #153

test/test-platform/TestPlatform.cpp
dawagner: ditto

PR #148, opened 2015-07-10
client simulator: Add CMake support

soaresm:
CMakeLists.txt was missing for the client simulator.

Signed-off-by: Michael Soares <michaelx.soares@intel.com>


soaresm:
@krocard and @dwagner please review

CMakeLists.txt
```diff
@@ -26,8 +26,13 @@
 # (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
 # SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
+option(CLIENT_SIMULATOR "Install the client simulator binary" OFF)
```
dawagner: Please make the client simulator a separate project (in the same way as the skeleton-subsystem). This will remove the need to add an option or any other change to this file.

CMakeLists.txt
```diff
@@ -26,8 +26,13 @@
...
+if (CLIENT_SIMULATOR)
+    cmake_minimum_required(VERSION 3.1)
```
dawagner: Travis has CMake 2.8; you can remove this.

CMakeLists.txt
dawagner: BTW, you'll need to edit .travis.yml in order to test the client simulator's installation. Note for later: have CI tests for the client simulator.

tools/clientSimulator/CMakeLists.txt
```diff
@@ -0,0 +1,51 @@
...
+    ${PYTHON_EXECUTABLE} -c
+        "from distutils import sysconfig;\\
+         print(sysconfig.get_python_lib(plat_specific=True, prefix=''))"
+    OUTPUT_VARIABLE PYTHON_MODULE_PATH OUTPUT_STRIP_TRAILING_WHITESPACE)
+
+install(DIRECTORY configuration DESTINATION ${PYTHON_MODULE_PATH})
+install(DIRECTORY criterion DESTINATION ${PYTHON_MODULE_PATH})
+install(DIRECTORY scenario DESTINATION ${PYTHON_MODULE_PATH})
+install(DIRECTORY testGenerator DESTINATION ${PYTHON_MODULE_PATH})
+install(DIRECTORY userInteraction DESTINATION ${PYTHON_MODULE_PATH})
```
dawagner: We need a slight code reorg: If you take a look at `/usr/lib/python*/dist-packages/`, you'll see that each Python module is contained in its own directory - with the exception of modules consisting of a single file.

I'd move all the folders above in a common directory e.g. `clientsimulator` (because that's the style recommended by PEP 8); then you only need to install that one directory.

dawagner: You'll certainly also need to modify the `import` clauses in `pfClientSimulator.py`

dawagner:
Thanks Michael :) just a "few" comments.

CMakeLists.txt
soaresm: pfClientSimulator is using python3 explicitly.
CMake 2.8 has a known bug concerning finding python when several version are installed on host :
http://www.cmake.org/Bug/print_bug_page.php?bug_id=13794
That's why ...

dawagner:
@soaresm ok, I got it. There is indeed a bug in older CMake's module for finding the python interpreter that makes impossible to get another version than the system's default one.

It is fixed is newer 2.8 CMake releases (I've tested 2.8.11). See https://github.com/dawagner/parameter-framework/tree/soaresm-cmake_clientSimulator

Please also remember to rename the python module folder from "clientSimulator" to "clientsimulator" (as specified by https://www.python.org/dev/peps/pep-0008/#package-and-module-names)

tools/clientSimulator/CMakeLists.txt
```diff
@@ -0,0 +1,52 @@
...
+# WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+# DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR
+# ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+# (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+# LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON
+# ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+# (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+# SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+# has been tested on 3.1 only - should not work on older versions
```
dawagner: outdated comment

.travis.yml
```diff
@@ -18,10 +18,13 @@ addons:
     apt:
         sources:
             - ubuntu-toolchain-r-test
+            - kalaris-cmake
         packages:
             - swig
             - valgrind
             - g++-4.8
+            - cmake
+            - python3-dev
```
dawagner: Please update the comments above to explain why we are installing (upgrading, in fact) these two packages.

PR #149, opened 2015-07-15
Check duplicate criterion values

dawagner:
- Revert the previously-merged patch for this issue: it wasn't fixing the root cause.
- Implement another solution that checks for and refuses duplicate criterion numerical values.


dawagner:
@clero please review

PR #150, opened 2015-07-17
Forward libxml2 error messages

dawagner:
Most error messages from libxml2 are missed or dropped. These changes makes the CXmlSerializingContext automatically gather them so that they will be logged or returned to the client.


dawagner:
Tested and approved internally.

PR #152, opened 2015-07-22
Add Floating point parameters

dawagner:
Add support for floating point parameters

The get/setAttribute inner API is reworked as preparative work.

Functional tests for this new parameter type are also added.
[...]

PR #154, opened 2015-07-23
Rework get/setAttribute

dawagner:
getAttribute API forces the user to call a specific API for each call
getAttributeString etc..  This patch replaces this API by a template one.  The
success of of the action can now be checked.  It can be useful one to check
that a conversion succeed.

Same for setAttribute* methods.

### DIFF
--- a/parameter/CompoundRule.cpp
+++ b/parameter/CompoundRule.cpp
@@ -140,7 +140,7 @@ bool CCompoundRule::matches() const
 bool CCompoundRule::fromXml(const CXmlElement& xmlElement, CXmlSerializingContext& serializingContext)
 {
     // Get type
-    _bTypeAll = xmlElement.getAttributeBoolean("Type", _apcTypes[true]);
+    _bTypeAll = xmlElement.getAttributeString("Type") == _apcTypes[true];
 
     // Base
     return base::fromXml(xmlElement, serializingContext);
--- a/parameter/CompoundRule.cpp
+++ b/parameter/CompoundRule.cpp
@@ -140,7 +140,9 @@ bool CCompoundRule::matches() const
 bool CCompoundRule::fromXml(const CXmlElement& xmlElement, CXmlSerializingContext& serializingContext)
 {
     // Get type
-    _bTypeAll = xmlElement.getAttributeString("Type") == _apcTypes[true];
+    string strType;
+    xmlElement.getAttribute("Type", strType);
+    _bTypeAll = strType == _apcTypes[true];
 
     // Base
     return base::fromXml(xmlElement, serializingContext);
@@ -150,7 +152,7 @@ bool CCompoundRule::fromXml(const CXmlElement& xmlElement, CXmlSerializingContex
 void CCompoundRule::toXml(CXmlElement& xmlElement, CXmlSerializingContext& serializingContext) const
 {
     // Set type
-    xmlElement.setAttributeString("Type", _apcTypes[_bTypeAll]);
+    xmlElement.setAttribute("Type", _apcTypes[_bTypeAll]);
 
     // Base
     base::toXml(xmlElement, serializingContext);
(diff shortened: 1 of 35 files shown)


dawagner:
This contains part of #152 which I will rebase once this PR is merged.

@krocard please review.

parameter/CompoundRule.cpp
```diff
@@ -140,7 +140,9 @@ bool CCompoundRule::matches() const
 bool CCompoundRule::fromXml(const CXmlElement& xmlElement, CXmlSerializingContext& serializingContext)
 {
     // Get type
-    _bTypeAll = xmlElement.getAttributeBoolean("Type", _apcTypes[true]);
+    string strType;
+    xmlElement.getAttribute("Type", strType);
+    _bTypeAll = strType == _apcTypes[true];
```
krocard: It would be so much nicer to returning the value. Maybe you could take a reference to a bool to warn on error.

dawagner: What value should be returned if getAttribute fails ?

If it returns the value of the default constructor, this leads to:

```
bool error;
string strType = xmlElement.getAttribute("Type", &error);
_bTypeAll = strType == _apcTypes[true];
```

is it so much nicer ?

parameter/EnumValuePair.cpp
```diff
@@ -80,10 +82,10 @@ void CEnumValuePair::logValue(string& strValue, CErrorContext& errorContext) con
 void CEnumValuePair::toXml(CXmlElement& xmlElement, CXmlSerializingContext& serializingContext) const
 {
     // Literal
-    xmlElement.setAttributeString("Literal", this->getName());
+    xmlElement.setAttribute("Literal", this->getName());
```
krocard: s/this->//

xmlserializer/XmlElement.cpp
```diff
@@ -271,3 +262,18 @@ bool CXmlElement::CChildIterator::next(CXmlElement& xmlChildElement)
     return false;
 }
 
+template bool CXmlElement::getAttribute(const std::string &name, std::string &value) const;
+template bool CXmlElement::getAttribute(const std::string &name, uint16_t &value) const;
+template bool CXmlElement::getAttribute(const std::string &name, uint32_t &value) const;
+template bool CXmlElement::getAttribute(const std::string &name, int32_t &value) const;
+template bool CXmlElement::getAttribute(const std::string &name, uint64_t &value) const;
```
krocard: I think you forget a couple:
 - bool
 - {unsigned,signed,} char
 - {unsigned,} short
 - {unsigned,} int
 - {unsigned,} long
 - {unsigned,} long long
 - float
 - {long,} double
 - std::string

xmlserializer/XmlElement.cpp
```diff
@@ -271,3 +262,18 @@ bool CXmlElement::CChildIterator::next(CXmlElement& xmlChildElement)
...
+template bool CXmlElement::getAttribute(const std::string &name, bool &value) const;
+template bool CXmlElement::getAttribute(const std::string &name, double &value) const;
+template bool CXmlElement::getAttribute(const std::string &name, float &value) const;
+
+template void CXmlElement::setAttribute(const std::string& name, const std::string &value);
+template void CXmlElement::setAttribute(const std::string& name, const bool &value);
+template void CXmlElement::setAttribute(const std::string& name, const int32_t &value);
+template void CXmlElement::setAttribute(const std::string& name, const uint32_t &value);
+template void CXmlElement::setAttribute(const std::string& name, const uint64_t &value);
+template void CXmlElement::setAttribute(const std::string& name, const float &value);
```
krocard: Why is there not a many as previously ?

Maybe you could do that in a loop. (that was a joke)

krocard:
:-1: Your first commit tittle is so long it does not fit in web interface.

xmlserializer/XmlElement.cpp
dawagner: bool, float, double and string are already covered. most integer types are covered too. Compiling the Parameter Framework does not necessitates more template instances than the one defined here. When this changes, they can be added.

krocard: true, enough. I found no nice way to get the list of supported types to instantiate them automatically.

PR #332, opened 2016-01-11
Allow setting boolean parameter with "true" and "false"

krocard:
Boolean parameters can now be set with values such as "true" and "false".


dawagner:
This pull request is now managed with CodeReviewHub. <a href='https://github.com/01org/parameter-framework/pull/332'>Refresh</a>.<a href='#crh-update-none'></a>

PR #338, opened 2016-01-15
Loose numeric format

krocard:
Allow non canonic numerical forms like:
 - 000 for 0
 - 0010 for 10
 - 10E3 for 10e3


dawagner:
numbers starting with 0 are usually octal representations.

PR #390, opened 2016-12-28
Hungarian notation string strError to string Error 

SnrNotHere16:
This is my attempt to resolve the Remove Hungarian Notation issue. I changed the identifier of variable string  strError to string Error. I also changed all the instances of strError to Error.

### DIFF
--- a/remote-process/main.cpp
+++ b/remote-process/main.cpp
@@ -42,19 +42,19 @@ using namespace std;
 
 bool sendAndDisplayCommand(asio::ip::tcp::socket &socket, CRequestMessage &requestMessage)
 {
-    string strError;
+    string Error;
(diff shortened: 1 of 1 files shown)


dawagner:
This pull request is now managed with CodeReviewHub. <a href='https://github.com/01org/parameter-framework/pull/390'>Refresh</a>.<a href='#crh-update-none'></a>

dawagner:
Hi, @SnrNotHere16. Thanks for tackling this issue.

Our coding style uses `camelCase` for variables and only uses `UpperCamelCase` for type names. It means that `strError` should become `error`. Also, it seems that you only updated `remote-process/main.cpp`; is that intended?

